Issue #9, opened 2020-05-05
Changing Password


samstronghammer:
If the user cancels or encounters an error during the new key creation within the "change lock" menu, the user needs to delete and redownload the app. This is a serious bug.

The reason for the bug is that the old secret key needs to be deleted before the new secret key is created if it is to retain the same tag in the secure enclave (to allow for consistent retrieval). A workaround would be to have an appended number to the tag that increases when the key is regenerated. This appended number can be kept track of dynamically in a separate file. Then the old key can be deleted _after_ the new key is generated, and not be deleted in case of error. Additionally, a message will need to be displayed to the user in case of error.

joaedwar:
@samstronghammer 

**You are invited to our Community Meeting!**
Tella community meetings provide a virtual platform for those interested in Tella to meet, get to know each other, and learn about the uses, applications, and needs of Tella. The event is scheduled for **Thursday, Oct 22 @ 2 GMT**. Click here to [add the invite](http://evt.mx/yadcOPPH) to your calendar. You can join using this [link](https://us02web.zoom.us/j/89895144942). Please be sure to also follow us out on [Twitter](https://twitter.com/Tella_app/status/1318084407757049858). Hope to see you there!

PR #26, opened 2020-08-13
Fix bug for Changing Password #9

rance-attack:
### Changes
- Append an UUID to application tag of private keys
- Store private/public key files in a directory named the UUID mentioned above
- Delete the old key after the new key is generated
- Remove usage of meta public key file
- Add a back button on `ChangeView`

### Known issues
Deleting key of **application password** might require entering the password again.
The worst case scenario is changing from password A to B.
1. Tap Setting > Change lock, enter A.
2. Tap Password, enter B twice.
3. Enter A for deleting the old key, this step is quite confusing.


Raph33:
This fixed the bug, thank you @rance-attack! We can now change the lock type without problem. 
cc @Anessa-Petteruti

PR #68, opened 2023-07-11
Feature/add snackbars

dhekra-rouatbi:

## Type of change

**Description:**

Display snackbars for report feature


Tella/Components/TrailingButtonToolbar.swift
```diff
@@ -0,0 +1,39 @@
...
+       
+        ToolbarItem(placement: .navigationBarTrailing) {
+           
+            Button {
+                completion?()
+            } label: {
+                Text(title)
+                .font(.custom(Styles.Fonts.regularFontName, size: 14))
+                .foregroundColor(Color.white)
+                .frame(width: 260,height:25,alignment:.trailing)
```
ValbuenaG, line 22: Are these values going to be always the same? I'm wondering if it would make sense to pass these values by params to make it more flexible in case we need it in the future.

Tella/Data/Networking/Upload/UploadService.swift
```diff
@@ -64,6 +63,23 @@ class UploadService: NSObject {
         activeOperations.append(operation)
         uploadQueue.addOperation(operation)
         operation.startUploadReportAndFiles()
+
+        // Display information toast
+        operation.response.sink { completion in
+        } receiveValue: { response in
+            switch response {
+            case .finish(let isAutoDelete , let title):
```
ValbuenaG, line 71: could we extract the logic into separate functions for better readability? Something like this:

```
func handleResponse(_ response: Response) {
    switch response {
    case .finish(let isAutoDelete, let title):
        displaySubmittedToast(title: title)
        if isAutoDelete {
            displayDeletedToast(title: title)
        }
    default:
        break
    }
}

func displaySubmittedToast(title: String?) {
    let message = String(format: LocalizableReport.reportSubmittedToast.localized, title ?? "")
    Toast.displayToast(message: message)
}

func displayDeletedToast(title: String?) {
    DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
        let message = String(format: LocalizableReport.reportDeletedToast.localized, title ?? "")
        Toast.displayToast(message: message)
    }
}

operation.response
    .sink(receiveCompletion: { _ in })
    { response in
        handleResponse(response)
    }
    .store(in: &subscribers)
    ```

**Review by ValbuenaG — COMMENTED** [review]
Great job! I think the PR looks great overall, I just added some minor suggestions

File: Tella/Components/TrailingButtonToolbar.swift (hunk 1 above)
Comment (dhekra-rouatbi) [reply Tella/Components/TrailingButtonToolbar.swift:22]: In this case I'll remove the width parameter and keep the height

**Review by rob-horizontal-iOS — APPROVED** [review]
Great work! I think it is PR is fine just make the changes as suggested by @ValbuenaG then we are good to go

File: Tella/Data/Database/TellaDataBase.swift [hunk 3]
```diff
@@ -509,24 +509,36 @@ class TellaDataBase {
...
+        let reportIds = submittedReports.compactMap{$0.id}
+
+        deleteReportFiles(reportIds: reportIds)
+
+        let reportCondition = [KeyValue(key: D.cStatus, value: ReportStatus.submitted.rawValue)]
+        
+        return try statementBuilder.delete(tableName: D.tReport,
+                                           primarykeyValue: reportCondition)
+    }
+
```
ahlem-jarrar, line 531: Maybe we should change the name to `deleteSubmittedReportsCount` as the function is also returning the number of deleted reports.

Tella/Data/Database/TellaDataBase.swift
```diff
@@ -509,24 +509,36 @@ class TellaDataBase {
...
+        deleteReportFiles(reportIds: [reportId])
+
         let reportCondition = [KeyValue(key: D.cReportId, value: report.id as Any)]
         return try statementBuilder.delete(tableName: D.tReport,
                                            primarykeyValue: reportCondition)
     }
-    
-    func deleteReportFiles(report:Report) {
+
+    func deleteSubmittedReport() throws -> Int {
```
ahlem-jarrar, line 519: Maybe instead of throwing an error from the function, we can catch the error within the function and handle it appropriately, with a `try catch`?

ahlem-jarrar (changes requested):

Tella/Data/Networking/Upload/UploadService.swift
dhekra-rouatbi: I did your suggestion and I centralised the code into a function.

ahlem-jarrar (approved):

PR #69, opened 2023-07-14
Gus/delete after fail - WIP

ValbuenaG:

## Delete files and connections after max failed attempts

User is now able to set a max failed attempts in the settings view. If the unlock attempts is equal than the max attempts should delete files and server connections

Fixes #issue-number

## Proposed changes 

- [x] Add delete after fail in sec settings
- [x] Move some sett security views into common to reuse in different parts of the code
- [x] Add DeleteAfterFailOption and DeleteAfterFailOptionsStatus into settingsModel
- [x] Add logic in LockViewModel to show the user a warning when there is 3 or less attempts before removing all data
- [x] Show message in both UnlockPasswordView and UnlockPinView
- [ ] Remove all files and server connections


Tella/Data/MainAppModel.swift
```diff
@@ -107,6 +107,14 @@ class MainAppModel: ObservableObject, AppModelFileManagerProtocol {
         vaultManager.removeAllFiles()
         publishUpdates()
     }
+
+    func deleteAllServersConnection() {
```
ValbuenaG, line 111: I move this method from serverModel to MainAppModel so we can use it in multiple places without the need of importing server model. Also is using `vaultManager` which corresponds to the appModel so I think it makes sense to be inside the appModel. 

Let me know what do you think @dhekra-rouatbi @RobertShrestha

Tella/Scenes/ContentView/ContentView.swift
```diff
@@ -19,17 +20,17 @@ struct ContentView: View {
...
-                    .environmentObject(LockViewModel(unlockType: .new))
+                    .environmentObject(LockViewModel(unlockType: .new, appModel: appModel))
             }
             
             if appViewState.currentView == .UNLOCK {
                 let passwordType = AuthenticationManager().getPasswordType()
                 passwordType == .tellaPassword ?
                 UnlockPasswordView()
-                    .environmentObject(LockViewModel(unlockType: .new))
+                    .environmentObject(LockViewModel(unlockType: .new, appModel: MainAppModel.stub()))
```
rob-horizontal-iOS, line 30: Is there any specific reason to app the stub() instead of appModel?
I think passing the appModel directly is also good is there any specific reason to generate a new instance of MainAppModel in both scenario

dhekra-rouatbi (changes requested):

Tella/Scenes/Authentication/ViewModel/LockViewModel.swift
```diff
@@ -74,7 +70,22 @@ class LockViewModel: ObservableObject {
     }
     
     func removeFilesAndConnections () -> Void {
-        appModel.removeAllFiles()
-        appModel.deleteAllServersConnection()
+        let fileManager = FileManager.default
```
dhekra-rouatbi, line 85: @ValbuenaG  Could you add a function in Vaultmanager that can delete the content of document folder 
You can use `containerURL` and `filemanager`

dhekra-rouatbi (changes requested):

Tella/Scenes/Authentication/Views/Unlock/UnlockView.swift
```diff
@@ -0,0 +1,162 @@
...
+        
+        .overlay(lockChoiceView)
+        .onAppear {
+            viewModel.initUnlockData()
+        }
+    }
+    
+    var titleString : String {
+        if(type == .tellaPin) {
+            if viewModel.shouldShowUnlockError {
```
dhekra-rouatbi, line 98: @ValbuenaG Could you put this code in a var that returns the pin or password title string and use switch case

dhekra-rouatbi (changes requested):

dhekra-rouatbi (changes requested):

Tella/Data/MainAppModel.swift
```diff
@@ -108,6 +132,26 @@ class MainAppModel: ObservableObject, AppModelFileManagerProtocol {
         publishUpdates()
     }
     
+    func deleteAfterMaxAttempts() {
+        let fileManager = FileManager.default
```
dhekra-rouatbi, line 136: @ValbuenaG "Could you add a function in Vaultmanager that can delete the content of document folder"

Yes you need to move the logic to vaultmanager because we have already the `containerURL` and we have `fileManager` that contains already `removeItem` function

dhekra-rouatbi (changes requested):

dhekra-rouatbi:
@ValbuenaG I think we need also to clear the settings after unlock fail

Tella/Scenes/Authentication/Views/Unlock/UnlockView.swift
```diff
@@ -0,0 +1,162 @@
...
+        if !viewModel.shouldShowUnlockError {
+            successLogin()
+        } else {
+            checkUnlockAttempts()
+        }
+    }
+    
+    private func successLogin() {
+        viewModel.unlockAttempts = 0
+        UserDefaults.standard.set(viewModel.unlockAttempts, forKey: "com.tella.lock.attempts")
```
dhekra-rouatbi, line 127: @ValbuenaG  Could you add the unlock Attempts to settings model

dhekra-rouatbi (changes requested):

Tella/Scenes/Authentication/Views/Unlock/UnlockView.swift
ValbuenaG: what would be the purpose of adding UnlockAttempts to the settings model? We are using this only on the LockModel for specific lock view behavior. Is it worth adding it to Settings? Just double checking in case I missed something

dhekra-rouatbi: In my opinion it cannot be added separately, to centralise the informations in one place

ValbuenaG:
@dhekra-rouatbi all of the comments are addressed except for "Could you add a function in Vaultmanager that can delete the content of document folder"

I move the logic to the vaultManager but I couldn't get the same result using the ContainerUrl Method that we already have. Maybe we can have a quick sync tomorrow to sort that out. 

Once we have that I'll fix the merge conflicts and update the rest of the translations files and we will good for release :)

PR #79, opened 2023-08-24
Uwazi/entity creation

ValbuenaG:

## Type of change

**Description:**

Fixes #issue-number

## Proposed changes 

*
*
*


Tella/Scenes/Uwazi/ViewModel/DraftUwaziEntity.swift
```diff
@@ -14,27 +14,77 @@ class DraftUwaziEntity: ObservableObject {
...
-            }
-        }
+        handlePdfsPrompt()
+        handleSupportPrompt()
+        handleTitlePrompt()
+        handleEntryPromptForProperties()
     }
+
+
+    fileprivate func handlePdfsPrompt() {
```
ValbuenaG, line 44: I think this shouldn't be inside the view model because it is mostly for parsing the different types of entity. Maybe moving it into a UwaziParser class

Tella/Scenes/Uwazi/ViewModel/DraftUwaziEntity.swift
```diff
@@ -14,27 +14,77 @@ class DraftUwaziEntity: ObservableObject {
...
+        let pdfPrompt = UwaziEntryPrompt(id: "10242050",
```
ValbuenaG, line 45: why are we using that id? Does all the pdf components have the same id?

Tella/Scenes/Uwazi/ViewModel/DraftUwaziEntity.swift
```diff
@@ -14,27 +14,77 @@ class DraftUwaziEntity: ObservableObject {
...
+                             required: $0.propertyRequired,
+                             helpText: $0.translatedLabel,
+                             selectValues: $0.values)
+
+        } ?? []
+        entryPrompts.append(contentsOf: entryPromptyProperties)
+    }
+
     
     func bindingForLabel(_ label: String) -> Binding<String> {
```
ValbuenaG, line 95: now that we are using this other approach for saving the data I think we can remove this one

ValbuenaG (commented):
I think the approach of saving data as entryPrompts is fine. Let's just move some of the parsers outside of the viewModel. 

Also, let's start adding the translations for the strings so because then it will be a lot of accumulated work

Tella/Scenes/Uwazi/Views/Widgets/UwaziTextWidget.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    var body: some View {
+        VStack(alignment: .leading) {
+                if mandatoryError {
+                    UwaziEntityMandatoryTextView()
+                }
+                TextfieldView(
+                    fieldContent: $value.stringValue,
+                    isValid: $isValidText,
+                    shouldShowError: .constant(false),
```
ValbuenaG, line 25: should this be always false?

Tella/Scenes/Uwazi/Views/Widgets/Common/UwaziEntityTitleView.swift
```diff
@@ -0,0 +1,40 @@
...
+
+struct UwaziEntityTitleView: View {
+    var title: String
+    var body: some View {
+        Group {
+            HStack {
+                Text(title)
+                    .font(.custom(Styles.Fonts.regularFontName, size: 14))
+                    .foregroundColor(Color.white)
+                Text("*")
```
ValbuenaG, line 19: this should be rendered conditionally only if the component is required. I think in some cases the component does not need this asterisk

Tella/Scenes/Uwazi/ViewModel/DraftUwaziEntity.swift
rob-horizontal-iOS: Its a initial draft but I was going to move it to a different class anyways
Thanks for the suggestion

Tella/Scenes/Uwazi/ViewModel/DraftUwaziEntity.swift
rob-horizontal-iOS: Like discussed it will have a clear explanation when we have a discussion with android team about it. I just used the implementation of Android.

Tella/Scenes/Uwazi/ViewModel/DraftUwaziEntity.swift
rob-horizontal-iOS: It is removed

Tella/Scenes/Uwazi/Views/Widgets/UwaziTextWidget.swift
rob-horizontal-iOS: According to the UI UX of this screen we are not using the common way of showing error i.e using shouldShowError rather than that we will be showing a text that show "This field is mandatory." below the title so it is set for as false permanently

Tella/Scenes/Uwazi/Views/Widgets/Common/UwaziEntityTitleView.swift
rob-horizontal-iOS: Ya as it is a initial draft so left it to only concentrate on the core thing. I have handled it.

dhekra-rouatbi (changes requested):

Tella/Scenes/Uwazi/Utils/Protocol/UwaziEntityParserProtocol.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Tella
+//
+//  Created by Gustavo on 29/09/2023.
+//  Copyright © 2023 HORIZONTAL. All rights reserved.
+//
+
+import Foundation
+
+protocol UwaziEntityParserProtocol {
+    var entryPrompts: [UwaziEntryPrompt] { get set }
```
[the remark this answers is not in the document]
ValbuenaG: The main reason was that we wanted to define a common interface with specific methods to parse the multiple types of data that Uwazi provides. At first, we started adding these methods in the UwaziEntityVM but I thought it was getting quite large and complex so we decided to abstract that logic into a protocol so we could re-use it in different parts of the uwazi flow.

Tella/Scenes/Uwazi/ViewModel/TemplateCardViewModel.swift
```diff
@@ -14,13 +14,15 @@ class TemplateCardViewModel: Hashable {
     var serverName: String
     var translatedName: String
     var deleteTemplate: (() -> Void)
+    var template: CollectedTemplate
```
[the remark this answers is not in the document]
ValbuenaG: Yeah i agree. The problem is that I need to use other props of template apart from the id (name, and properties). What do you think if we add properties and commonProperties to templateCardViewModel and just pass the templateCardViewModel like this:

```
 navigateTo(destination: CreateEntityView(template: templateCardViewModel).environmentObject(sheetManager))
                                    sheetManager.hide()
```

PR #81, opened 2023-08-25
Fix/delete after fail fixes

ValbuenaG:

## Type of change

**Description:**

Fixes #issue-number

## Proposed changes 

* Fix copies of "off" selection in delete after fail bottom sheet
* Add "show warning attempts" toggle
* Add snackbars

### DIFF
--- a/Tella/Scenes/Settings/ViewModel/SettingsViewModel.swift
+++ b/Tella/Scenes/Settings/ViewModel/SettingsViewModel.swift
@@ -69,6 +69,12 @@ class SettingsViewModel: ObservableObject {
     }
     
     func saveDeleteAfterFail() {
+        if(appModel.settings.deleteAfterFail == .off && selectedDeleteAfterFailOption != .off) {
+            appModel.settings.showUnlockAttempts = true
+        }
+        if(selectedDeleteAfterFailOption == .off) {
+            appModel.settings.showUnlockAttempts = false
+        }
         appModel.settings.deleteAfterFail = selectedDeleteAfterFailOption
         appModel.saveSettings()
     }
--- a/Tella/Scenes/Settings/ViewModel/SettingsViewModel.swift
+++ b/Tella/Scenes/Settings/ViewModel/SettingsViewModel.swift
@@ -69,14 +69,18 @@ class SettingsViewModel: ObservableObject {
     }
     
     func saveDeleteAfterFail() {
+        setShowUnlockAttempts()
+        appModel.settings.deleteAfterFail = selectedDeleteAfterFailOption
+        appModel.saveSettings()
+    }
+    
+    func setShowUnlockAttempts () {
         if(appModel.settings.deleteAfterFail == .off && selectedDeleteAfterFailOption != .off) {
             appModel.settings.showUnlockAttempts = true
         }
         if(selectedDeleteAfterFailOption == .off) {
             appModel.settings.showUnlockAttempts = false
         }
-        appModel.settings.deleteAfterFail = selectedDeleteAfterFailOption
-        appModel.saveSettings()
     }
     
     func cancelDeleteAfterFail() {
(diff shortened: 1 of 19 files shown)


dhekra-rouatbi:
The explanation of `Show unlock attempts remaining` message is truncated 

IMG_3579

Tella/Scenes/Settings/ViewModel/SettingsViewModel.swift
```diff
@@ -69,6 +69,12 @@ class SettingsViewModel: ObservableObject {
     }
     
     func saveDeleteAfterFail() {
+        if(appModel.settings.deleteAfterFail == .off && selectedDeleteAfterFailOption != .off) {
```
dhekra-rouatbi, line 72: or simply you can write 

        if selectedDeleteAfterFailOption != .off {
            appModel.settings.showUnlockAttempts = true
        }
       
 because in the case of  `selectedDeleteAfterFailOption` equal to `off` we don't check `showUnlockAttempts`
        
or 
       
 simply  `appModel.settings.showUnlockAttempts = selectedDeleteAfterFailOption != .off ? true : false`

Tella/Scenes/Settings/Views/Security/DeleteAfterFailView.swift
```diff
@@ -36,6 +36,14 @@ struct DeleteAfterFailView: View {
             }, cancelLabel: LocalizableSettings.settLockTimeoutCancelSheetAction.localized, saveAction: {
                 settingsViewModel.saveDeleteAfterFail()
                 sheetManager.hide()
+                
+                let message: String
```
dhekra-rouatbi, line 40: Could you put this code into another function for example `displayToast`

Tella/Scenes/Settings/Views/Security/DeleteAfterFailView.swift
```diff
@@ -36,6 +36,14 @@ struct DeleteAfterFailView: View {
             }, cancelLabel: LocalizableSettings.settLockTimeoutCancelSheetAction.localized, saveAction: {
                 settingsViewModel.saveDeleteAfterFail()
                 sheetManager.hide()
+
```
dhekra-rouatbi, line 39: I think it's better to move the code of `BottomButtonsView` into a separate variable that returns this view

dhekra-rouatbi (changes requested):

dhekra-rouatbi:
I think the unlock attempts are not saved because I have 5 attempts, I tried to unlock twice then I close the app and I relaunch the app, the number of attempts are 5 not 3

Tella/Scenes/Settings/Views/Security/DeleteAfterFailView.swift
ValbuenaG: You are right! The view got a little bit more complex after adding the display toast logic. I isolated the BottomButtonsView and created a function for the toast to avoid the over complexity in one view

Tella/Scenes/Settings/ViewModel/SettingsViewModel.swift
ValbuenaG: Yeah I thought of this at the beginning but there is this edge case:

- User has delete after failed set to 5 attempts
- User switch toggle to off
- User change from 5 attempts to 10 attempts

In this case I think the toggle just need to remain off. And with this logic: `appModel.settings.showUnlockAttempts = selectedDeleteAfterFailOption != .off ? true : false` It will be toggle to true everytime the user change the number of attempts. 


I will double-check with Caro and Yann Chwenn to know what the correct behavior is.

ValbuenaG:
> I think the unlock attempts are not saved because I have 5 attempts, I tried to unlock twice then I close the app and I relaunch the app, the number of attempts are 5 not 3

good catch! It was a missing declaration on the SettingsModel. I just fixed it

Tella/Scenes/Settings/Views/Security/SecuritySettingsView.swift
```diff
@@ -75,16 +75,31 @@ struct SecuritySettingsView: View {
...
+            }
+        }
+    }
+    
     var deleteAfterFailView: some View {
         
         SettingsItemView<AnyView>(imageName: "settings.lock",
                                   title: LocalizableSettings.settSecDeleteAfterFail.localized,
-                                  value: appModel.settings.deleteAfterFail.displayName,
+                                  value: showDeleteAfterFailSelected(),
```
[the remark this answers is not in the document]
ValbuenaG: Oh I didn't think about this. Thanks for the suggestion!

dhekra-rouatbi (approved):

PR #88, opened 2023-09-15
Character limit in Uwazi login screen fields

rob-horizontal-iOS:

## Type of change
Remove the limitation of character in Uwazi login's username and password textfield so that it will enable the login button when at least 1 character each is inserted in both field.

- [x] Bug fix *(Fixes an issue)*

## Proposed changes 

- Added validator for uwazi username and password validation
- Added new enum uwaziUsername and uwaziPassword for validation 
- Change the fieldType of username and password to uwaziUsername and uwaziPassword

### DIFF
--- a/Tella/Components/TextfieldView.swift
+++ b/Tella/Components/TextfieldView.swift
@@ -11,8 +11,6 @@ enum FieldType {
     case text
     case password
     case code
-    case uwaziUsername
-    case uwaziPassword
 }
 struct TextfieldView : View {
     
@@ -157,10 +155,6 @@ struct TextfieldView : View {
             
         case .code:
             self.isValid = value.codeValidator()
-        case .uwaziUsername:
-            self.isValid = value.validateWithRegex(pattern: Regex.textRegex)
-        case .uwaziPassword:
-            self.isValid = value.validateWithRegex(pattern: Regex.textRegex)
         }
         self.shouldShowError = false
     }
--- a/Tella/Components/TextfieldView.swift
+++ b/Tella/Components/TextfieldView.swift
@@ -5,7 +5,15 @@
 
 import SwiftUI
 import Combine
-
+enum FieldType {
+    case url
+    case username
+    case text
+    case password
+    case code
+    case uwaziUsername
+    case uwaziPassword
+}
 struct TextfieldView : View {
     
     @Binding var fieldContent : String
@@ -149,6 +157,10 @@ struct TextfieldView : View {
             
         case .code:
             self.isValid = value.codeValidator()
+        case .uwaziUsername:
+            self.isValid = value.uwaziUsernameValidator()
+        case .uwaziPassword:
+            self.isValid = value.uwaziPasswordValidator()
         }
         self.shouldShowError = false
     }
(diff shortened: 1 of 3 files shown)


Tella/Utils/Validator.swift
```diff
@@ -74,6 +74,24 @@ extension String {
         }
         return true
     }
+    func uwaziUsernameValidator() -> Bool {
```
ValbuenaG, line 77: Since both `uwaziUsernameValidator` and `uwaziPasswordValidator ` are using the same logic. I think we could unify these two methods in one like this:

```
func validateWithRegex(pattern: String) -> Bool {
    guard !self.isEmpty else {
        return false
    }
    return validateRegex(value: self, pattern: pattern)
}
```

ValbuenaG (changes requested):

Tella/Utils/Validator.swift
rob-horizontal-iOS: Ya that makes scene but I think we should only use this function `validateWithRegex ` in the extension and remove all other validators from the extension and just place it into `viewmodel` where it is used so it is better for testability and good for readability.

As this branch is for the uwazi login validation I have not change the code as I explained but I have combined it into one function for now 

```Swift
func uwaziLoginValidtor() -> Bool {
        guard !self.isEmpty else {
            return false
        }
        guard validateRegex(value: self, pattern:  Regex.textRegex) else {
            return false
        }
        return true
    }
```

ValbuenaG (approved):

Tella/Utils/Validator.swift
```diff
@@ -74,6 +74,23 @@ extension String {
...
+    func validateWithRegex(pattern: String) -> Bool {
+        guard !self.isEmpty else {
+            return false
+        }
+        guard validateRegex(value: self, pattern: pattern) else {
+            return false
+        }
+        return true
+    }
+    func uwaziLoginValidtor() -> Bool {
```
dhekra-rouatbi, line 86: I think we have already this validator which is `textValidator` you can use it instead

Tella/Components/TextfieldView.swift
```diff
@@ -5,7 +5,15 @@
...
 import SwiftUI
 import Combine
-
+enum FieldType {
+    case url
+    case username
+    case text
+    case password
+    case code
+    case uwaziUsername
```
dhekra-rouatbi, line 14: I think we have already text field type that can be used for  uwaziUsername and  uwaziPassword

dhekra-rouatbi (changes requested):

dhekra-rouatbi:
I think this case is not treated `The login button should be enabled if an only iff both fields have some content`

rob-horizontal-iOS:
> I think this case is not treated `The login button should be enabled if an only iff both fields have some content`

I think it is working

Tella/Utils/Validator.swift
```diff
@@ -74,6 +74,14 @@ extension String {
         }
         return true
     }
-
+    func validateWithRegex(pattern: String) -> Bool {
```
dhekra-rouatbi, line 77: I think this function `validateWithRegex` is not used

dhekra-rouatbi (changes requested):

PR #104, opened 2023-10-25
Feature/feedback method

dhekra-rouatbi:
## Type of change

**Description:**


Tella/Data/Database/TellaDataBase.swift
```diff
@@ -591,3 +592,89 @@ class TellaDataBase {
...
+    
+    func getDraftFeedback() -> Feedback? {
+        do {
+            let feedbackCondition = [KeyValue(key: D.cStatus, value: FeedbackStatus.draft.rawValue)]
+
+            let feedbackDict = try statementBuilder.getSelectQuery(tableName: D.tFeedback,
+                                                                   equalCondition: feedbackCondition)
+            return try feedbackDict.first?.decode(Feedback.self)
+        } catch {
+            return nil
```
ValbuenaG, line 619: Could we add a Debug.Log error message in the catch blocks of these methods?

ValbuenaG, line 619: Could we add an error log message here? same in the rest of the methods

Tella/Data/Database/TellaDataBase.swift
```diff
@@ -591,3 +592,89 @@ class TellaDataBase {
...
+                                        primarykeyValue: feedbackCondition)
+            
+        } catch(let error) {
+            debugLog(error)
+        }
+        
+    }
+    
+    
+    func deleteFeedback(feedbackId: Int?) {
```
ValbuenaG, line 671: here we are sending feedbackId as optional so I supposed if the feedbackId is nil the query will crash right? Probably we should safely unwrap optional values before executing queries using if let or guard let.

```
guard let feedbackId = feedbackId else {
            return 
        }
 ```

File: Tella/Data/Networking/Repositories/FeedbackRepository.swift [hunk 3]
```diff
@@ -0,0 +1,74 @@
...
+        switch self {
+        case .submitFeedback:
+            return HTTPMethod.post
+        }
+    }
+    
+    var headers: [String : String]? {
+        switch self {
+        case .submitFeedback:
+            return [HTTPHeaderField.tellaPlatform.rawValue:"wearehorizontal"]
```
ValbuenaG, line 66: could we add "wearehorizontal" and "https://api.feedback.tella-app.org" as constants variables?

Tella/Domain/Entity/Feedback/Feedback.swift
```diff
@@ -0,0 +1,46 @@
...
+    var text: String?
+    var status : FeedbackStatus?
+    var createdAt, updatedAt: Date?
+    
+    enum CodingKeys: String, CodingKey {
+        case id = "c_id"
+        case text = "c_text"
+        case status = "c_status"
+        case createdAt = "c_created_date"
+        case updatedAt = "c_upated_date"
```
ValbuenaG, line 19: here is a typo. It should be  `case updatedAt = "c_updated_date"`

ValbuenaG (commented):
Looks good to me!  I just left a few comments with general suggestions and fixed some typos. Excellent work!

Tella/Data/Database/TellaDataBase.swift
dhekra-rouatbi: This won't crash but it's better to unwrap optional values before executing queries

PR #110, opened 2023-12-13
Background file encryption 

dhekra-rouatbi:
## Type of change

**Description:**


Tella/Components/TopSheetView/TopSheetView.swift
```diff
@@ -0,0 +1,130 @@
...
+                if drag > 0 {
+                    self.currentHeight = kmaxDrag
+                }
+                
+                self.currentDrag = .zero
+            })
+    }
+    
+    private func fraction_progress(lowerLimit: Double = 0, upperLimit:Double, current:Double, inverted:Bool = false) -> Double{
+        var val:Double = 0
```
ValbuenaG, line 110: what do you think of using a clamp function for getting the fraction of the value?

```
let clampedCurrent = min(max(current, lowerLimit), upperLimit)
let fraction = (clampedCurrent - lowerLimit) / (upperLimit - lowerLimit)
```

Here we are clamping the value to be within the lower and upper limit and then calculating the fraction using the clamp value.

then we could just return this:


```
return inverted ? fraction : 1 - fraction
```

Tella/Data/Networking/Feedback/EncryptionService/EncryptionOperation.swift
```diff
@@ -0,0 +1,53 @@
...
+    private var cancellable: AnyCancellable?
+    private var subscribers = Set<AnyCancellable>()
+    
+    init(mainAppModel :MainAppModel) {
+        super.init()
+        self.mainAppModel = mainAppModel
+    }
+    
+    override func main() {
+        super.main()
```
ValbuenaG, line 22: what is the purpose of calling super.main() here?

ValbuenaG (commented):
great work! I just have some small comments

Tella/Data/Crypto/Cryptor.swift
```diff
@@ -0,0 +1,161 @@
...
+        
+        if operationResult !=  kCCSuccess {
+            debugLog("Error encryption \(operationResult)", space: .crypto)
+            throw CryptoError.cryptorCreationFailed
+        }
+    }
+    
+    private func cryptorUpdate() throws {
+        var stop = false
+
```
ValbuenaG, line 87: could we add a defer block to close the inputFile and outputFile when the method exits? I know we are doing this inside of the while loop but this way the method is more readable

Tella/Data/Crypto/Cryptor.swift
```diff
@@ -0,0 +1,161 @@
...
+                        CCCryptorUpdate(
+                            cryptor!,
+                            inputBytes.baseAddress, inputBytes.count,
+                            outputBytes.baseAddress, outputBytes.count,
+                            &encryptedByteCount
+                        )
+                    }
+                }
+                
+                if operationResult != kCCSuccess {
```
ValbuenaG, line 118: if we add a defer we could get rid of the closeFile statements here and change this into a guard statement

```
guard operationResult == kCCSuccess else {
  throw CryptoError.encryptionFailed
}
```

Tella/Data/Crypto/Cryptor.swift
```diff
@@ -0,0 +1,161 @@
...
+                
+                // Read a chunk from the input file
+                
+                let inputBuffer = inputFile.readData(ofLength: Self.bufferSize) as Data
+                
+                var encryptedData = Data(count: Int(Self.bufferSize) + kCCBlockSizeAES128)
+                
+                if inputBuffer.isEmpty {
+                    // Reached end of file
+                    stop = true
```
ValbuenaG, line 101: Instead of having an stop variable and checking it to exit the while loop. Why don't we just add a break statement here? something simple as this: `guard !inputBuffer.isEmpty else { break }`

If we need to execute the rest of the code for this last part of the file we can just move it to the end of the loop right?

Tella/Data/Crypto/Cryptor.swift
```diff
@@ -0,0 +1,161 @@
...
+        
+        let finalStatus = finalData.withUnsafeMutableBytes { finalData in
+            CCCryptorFinal(
+                cryptor!,
+                finalData.baseAddress, finalData.count,
+                &bytesEncrypted
+            )
+        }
+        
+        if finalStatus != kCCSuccess  {
```
ValbuenaG, line 148: we can implement the same defer pattern and avoid closing the files here

Tella/Data/Crypto/Cryptor.swift
```diff
@@ -0,0 +1,161 @@
...
+            inputFile.closeFile()
+            outputFile.closeFile()
+            throw CryptoError.finalizationFailed
+        }
+        
+        CCCryptorRelease(cryptor!)
```
ValbuenaG, line 154: this line could also be added to the defer block

Tella/Data/Networking/Feedback/EncryptionService/EncryptionService.swift
```diff
@@ -0,0 +1,74 @@
...
+    func addEncryptionOperations( fileDetails:[VaultFileDetails], parentId: String?, shouldReloadVaultFiles:Binding<Bool>?) {
+        
+        for fileDetail in fileDetails {
+            operationQueue.addOperation({
+               
+                let operation = EncryptionOperation(mainAppModel: self.mainAppModel)
+                
+                operation.addVaultFile(fileDetail: fileDetail, filePath: fileDetail.fileUrl, parentId: parentId, mainAppModel: self.mainAppModel)?
+                    .receive(on: DispatchQueue.main)
+                    .sink(receiveValue: { backgroundResult in
```
ValbuenaG, line 65: should we also handle failures here?

ValbuenaG (changes requested):
Looks good! Just added a few comments

Tella/Data/Crypto/Cryptor.swift
dhekra-rouatbi: I think we need to put the `defer` in `cryptFile()` function because we still use  `inputFile` and `outputFile` in `cryptorFinal()` function

Tella/Data/Crypto/Cryptor.swift
dhekra-rouatbi: I think `break` doesn't work with `autoreleasepool` that's why I added a `stop` var

Tella/Data/Crypto/CryptoManager.swift
```diff
@@ -316,14 +323,62 @@ extension CryptoManager: CryptoManagerInterface {
...
+                debugLog("Failed to recover public key", space: .crypto)
+                return false
+            }
+            
+            guard let encryptionKeyData = publicKey.getData() else {
+                debugLog("Failed to recover private key data", space: .crypto)
+                return false
+            }
+
+            let crypt = try Cryptor(inputFileURL: inputFileURL, outputFileURL: outputFileURL, encryptionKeyData: encryptionKeyData, cryptoOperation: .encrypt)
```
ahlem-jarrar, line 347: For instance, consider renaming `crypt` to something like `fileCryptor`.

Tella/Data/Crypto/CryptoManager.swift
```diff
@@ -316,14 +323,62 @@ extension CryptoManager: CryptoManagerInterface {
...
-
+        
         guard let metaPrivateKey = self.metaPrivateKey else {
             debugLog("Failed to recover private key", space: .crypto)
             return nil
         }
         return decrypt(data, metaPrivateKey)
     }
     
+    func encryptFilePath(inputFileURL: URL, outputFileURL: URL) -> Bool {
```
ahlem-jarrar, line 334: let's add comments or documentation to explain the purpose of the function, parameters, and any potential side effects?

Tella/Data/Crypto/CryptoManager.swift
```diff
@@ -316,14 +323,62 @@ extension CryptoManager: CryptoManagerInterface {
...
+                debugLog("Failed to recover private key", space: .crypto)
+                return false
+            }
+            
+            guard let encryptionKeyData = metaPrivateKey.getData() else {
+                debugLog("Failed to recover private key data", space: .crypto)
+                return false
+            }
+
+            let crypt = try Cryptor(inputFileURL: inputFileURL, outputFileURL: outputFileURL, encryptionKeyData: encryptionKeyData, cryptoOperation: .decrypt)
```
ahlem-jarrar, line 371: Let's `Separate Concerns` : Separate the concerns of recovering the private key and decrypting the file?

ahlem-jarrar: maybe a function to decrypt the file like this : 
`private func decryptFile(inputFileURL: URL, outputFileURL: URL, encryptionKeyData: Data) throws {}`

ahlem-jarrar: and a function -> is now named `decryptFile(at:outputTo:)` for clarity.


`func decryptFile(at inputFileURL: URL, outputTo outputFileURL: URL) -> Bool {}`

Tella/Data/Crypto/CryptoManager.swift
```diff
@@ -316,14 +323,62 @@ extension CryptoManager: CryptoManagerInterface {
...
+            try crypt.cryptFile()
+
+            return true
+            
+        }
+        catch let error {
```
ahlem-jarrar, line 377: maybe use `decryptionError` instead to be more specific name convention?

Tella/Data/Crypto/Cryptor.swift
```diff
@@ -0,0 +1,165 @@
...
+// Enum to handle encryption errors
+enum CryptoError: Error {
+    case cryptorCreationFailed
+    case encryptionFailed
+    case finalizationFailed
+}
+
+class Cryptor {
+    
+    static let bufferSize = 1024 * 1024 * 8  // 8 MB buffer size
```
ahlem-jarrar, line 20: I have a question regrading the buffer size do we know why we are using this exact size ? because it depends on the nature of the Data -> If our data is composed of small chunks, using a smaller buffer size might be more efficient. On the other hand, if our data is typically larger, a larger buffer size could be beneficial.

Tella/Data/Database/Vault/VaultDataBase.swift
```diff
@@ -146,7 +163,39 @@ class VaultDatabase : DataBase, VaultDataBaseProtocol {
             return nil
         }
     }
+
+    func getNonUpdatedVaultFiles() -> [VaultFileDB] {
+        do {
+            let vaultFilesDict = try statementBuilder.getSelectQuery(tableName: VaultD.tVaultFile,
+                                                                     equalCondition:[KeyValue(key: VaultD.cEncryptionUpdated, value: false)])
+            
+            let vaultFiles = vaultFilesDict.compactMap{VaultFileDB.init(dictionnary: $0)}
```
ahlem-jarrar, line 172: maybe let's introduce  a private method `createVaultFile` to handle the creation of VaultFileDB objects? -> allowing for better separation of concerns?

Tella/Data/Database/Vault/VaultDataBase.swift
```diff
@@ -146,7 +163,39 @@ class VaultDatabase : DataBase, VaultDataBaseProtocol {
...
+            return vaultFiles
+            
+        } catch {
+            return []
```
ahlem-jarrar, line 176: Let's provide more meaningful error messages for better debugging, let's not keep this epmty ?

Tella/Data/Database/Vault/VaultDataBase.swift
```diff
@@ -146,7 +163,39 @@ class VaultDatabase : DataBase, VaultDataBaseProtocol {
...
+        }
+    }
     
+    @discardableResult
+    func updateEncryptionVaultFile(id: String?) -> Result<Bool, Error> {
```
ahlem-jarrar, line 181: `id` is nullable -> let use guard to ensure that the id is not nil, throwing a meaningful error if it is. what do you think?

Tella/Data/Database/Vault/VaultDataBase.swift
```diff
@@ -146,7 +163,39 @@ class VaultDatabase : DataBase, VaultDataBaseProtocol {
...
+        
+        do {
+            
+            let valuesToUpdate = [KeyValue(key: VaultD.cEncryptionUpdated, value: 1)]
+            let vaultCondition = [KeyValue(key: VaultD.cId, value: id)]
```
ahlem-jarrar, line 186: Let's not use magic numbers for the encryption updated value 1?

ahlem-jarrar (changes requested):

Tella/Data/Database/Vault/VaultDataBase.swift
dhekra-rouatbi: or maybe an array extension like this:

```
extension Array where Element == Dictionary<String,Any>{
      func getVaultFiles() -> [VaultFileDB] {
        return self.compactMap{VaultFileDB.init(dictionnary: $0)}
    }
}
```

PR #114, opened 2024-01-08
add scroll view in EditSettingsServerView

ValbuenaG:
This PR adds a ScrollView in the editSettingsServerView because it was getting cutoff in small screens.

### DIFF
--- a/Tella/Scenes/Settings/Views/Servers/EditSettingsServerView.swift
+++ b/Tella/Scenes/Settings/Views/Servers/EditSettingsServerView.swift
@@ -19,28 +19,29 @@ struct EditSettingsServerView: View {
     var body: some View {
         
         ContainerView {
-            
-            VStack {
-                
-                editServerHeaderView
-                
-                SettingsCardView(cardViewArray: [serverNameView.eraseToAnyView(),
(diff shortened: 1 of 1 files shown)


Tella/Scenes/Settings/Views/Servers/EditSettingsServerView.swift
```diff
@@ -19,28 +19,29 @@ struct EditSettingsServerView: View {
...
-                ])
-                
-                
-                Spacer()
-                
-                bottomView
+            ScrollView {
+                VStack {
+                    
+                    editServerHeaderView
```
dhekra-rouatbi, line 25: In my opinion, it's better to exclude the `editServerHeaderView` from the  `ScrollView`

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/Views/Servers/EditSettingsServerView.swift
ValbuenaG: you are right! I also excluded the bottom view so the only thing that would be scrollable is the content!

PR #118, opened 2024-01-30
Change title when removing a folder

ValbuenaG:
CHange title when removing a folder
## Type of change

**Description:**

- [x] Bug fix *(Fixes an issue)*

Fixes #issue-number

## Proposed changes 

* If user remove a folder it should show "delete folder?" or "delete folders?" instead of showing the message of the file
* Change file message to also show the amount of files to be removed
*

### DIFF
--- a/Tella/Scenes/HomeView/ViewModel/FileListViewModel.swift
+++ b/Tella/Scenes/HomeView/ViewModel/FileListViewModel.swift
@@ -391,19 +391,10 @@ extension FileListViewModel {
 }
 
 extension FileListViewModel {
-    
-    func getTotalFilesInFolder(folder: VaultFileDB) -> Int {
-        let allFilesInFolder = self.appModel.vaultFilesManager!.getVaultFile(vaultFilesFolders: [folder])
-
-        return allFilesInFolder.filter { $0.type != .directory }.count
-    }
-
     var deleteConfirmation: DeleteConfirmation {
         let selectedFolders = selectedFiles.filter { $0.type == .directory }
         let fileCount = selectedFiles.count - selectedFolders.count
-        let totalFilesInsideFolders = selectedFolders.reduce(0) { count, folder in
-            count + getTotalFilesInFolder(folder: folder)
-        }
+        let  totalFilesInsideFolders = self.appModel.vaultFilesManager!.getVaultFile(vaultFilesFolders: selectedFolders).count
 
         let selectionCountDetails = SelectionCountDetails(fileCount: fileCount,
                             foldersCount: selectedFolders.count,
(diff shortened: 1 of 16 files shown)


Tella/Scenes/HomeView/ViewModel/FileListViewModel.swift
```diff
@@ -389,3 +389,62 @@ extension FileListViewModel {
...
+        }
+    }
+
+    var deleteConfirmationMessage: String {
+        let selectedFolders = selectedFiles.filter { $0.type == .directory }
+        let folderCount = selectedFiles.filter { $0.type == .directory }.count
+        let fileCount = selectedFiles.count - folderCount
+        
+        let totalFilesInsideFolders = selectedFolders.reduce(0) { count, folder in
+            count + getFilesInFolder(parentId: folder.id ?? "")
```
dhekra-rouatbi, line 422: I don't think that this implementation will give you all the files under the selected folders, it will give you only the first files under the folder
I suggest to use `FileWalker` to get all the files, it's used in 
`func deleteVaultFile(vaultFiles : [VaultFileDB]) -> Result<Bool, Error>?`

ValbuenaG: ohh right, i forgot about nested folders. I will implement this rn!

Tella/Scenes/HomeView/ViewModel/FileListViewModel.swift
```diff
@@ -389,3 +389,27 @@ extension FileListViewModel {
         return FileListViewModel(appModel: MainAppModel.stub(), filterType: .all, rootFile: VaultFileDB.stub())
     }
 }
+
+extension FileListViewModel {
+    
+    func getTotalFilesInFolder(folder: VaultFileDB) -> Int {
+        let fileWalker = FileWalker(vaultDatabase: self.appModel.vaultFilesManager!.vaultDataBase)
```
dhekra-rouatbi, line 396: I think it's better to add this function in `VaultFilesManager`  

    func getVaultFile(vaultFilesFolders : [VaultFileDB]) -> [VaultFileDB] {
        var resultFiles : [VaultFileDB] = []
       let fileWalker = FileWalker(vaultDatabase: self.vaultDataBase)
        
        vaultFilesFolders.forEach { file in
            resultFiles.append(contentsOf: fileWalker.walk(root: file))
         }
        return resultFiles
    }



and `walk` function to get only the files in `FileWalker` class


    func walk(root:VaultFileDB) -> [VaultFileDB] {
        
        let vaultFiles = vaultDatabase?.getVaultFiles(parentId: root.id, filter: .all, sort: nil)
        
        vaultFiles?.forEach({ vaultFile in

            if vaultFile.type == .directory {
                _ = self.walk(root: vaultFile)
            } else {
                resultFiles.append(vaultFile)
            }
        })
        return resultFiles
    }

dhekra-rouatbi (changes requested):

dhekra-rouatbi (changes requested):

PR #119, opened 2024-02-01
first draft of resources ui

ValbuenaG:

## Type of change

**Description:**
This should be the base branch for resources where all others PRs will be merge here before merging with develop. 

Fixes #issue-number

## Proposed changes 

* Implement UI of resources
* Add view model and first draft of architecture
* Add localization


Tella/Components/ReloadButton.swift
```diff
@@ -0,0 +1,37 @@
...
+//  ReloadButton.swift
+//  Tella
+//
+//  Created by gus valbuena on 1/31/24.
+//  Copyright © 2024 HORIZONTAL. All rights reserved.
+//
+
+import SwiftUI
+
+struct ReloadButton: View {
```
dhekra-rouatbi, line 11: I think it's better to create a ReloadButton with ToolbarContent type instead to create another reloadButton function in View extension 


   ```
struct ReloadButtonn: ToolbarContent {
    var action: () -> Void
    var body: some ToolbarContent {
        ToolbarItem(placement: .navigationBarTrailing) {
            Button(action: action) {
                Image("arrow.clockwise")
                    .resizable()
                    .frame(width: 24, height: 24)
            }
        }
    }
 }
```

**Review by dhekra-rouatbi — CHANGES_REQUESTED** [review]

File: Tella/Scenes/Resources/ViewModel/ResourcesViewModel.swift [hunk 2]
```diff
@@ -0,0 +1,20 @@
...
+//
+
+import Foundation
+
+class ResourcesViewModel: ObservableObject {
+    @Published var downloadedResourcesVM : DownloadedResourcesVM
+    @Published var availableResourcesVM : AvailableResourcesVM
+    
+    init() {
+        self.downloadedResourcesVM = DownloadedResourcesVM()
```
dhekra-rouatbi, line 16: I think it's better to create one view model `ResourcesViewModel` that contains the code in the DownloadedResourcesVM and  AvailableResourcesVM because they are related and avoid passing in parameter later the downloadedResourcesVM to AvailableResourcesVM

dhekra-rouatbi (changes requested):

dhekra-rouatbi:
There is a wired animation in title 

https://github.com/Horizontal-org/Tella-iOS/assets/91288060/6b1e4e3e-aa01-44ca-8026-115ee2bccdb3

Tella/Supporting Files/en.lproj/Localizable.strings
```diff
@@ -377,3 +377,12 @@ This feedback is anonymous, so make sure to include contact information if you w
 "Settings_SettFeedback_ExitFeedback_SheetExpl" = "You can come back anytime to submit your feedback.";
 "Settings_SettFeedback_ExitFeedback_Exit_SheetAction" = "Exit without saving";
 "Settings_SettFeedback_ExitFeedback_Save_SheetAction" = "Save draft";
+
+"Resources_Home_Serve_title" = "Resources";
+"Resources_Downloaded_Title" = "Downloaded";
+"Resources_Downloaded_Empty" = "You have not downloaded any resources. Tap on the “download” icon next to each resource below to get started.";
+"Resources_Downloaded_Empty_SecondMsg" = "You have not downloaded any resources.";
```
dhekra-rouatbi, line 384: `Resources_Downloaded_Empty_SecondMsg` should be used if there are no resources downloaded and there are no resources to download

dhekra-rouatbi (changes requested):

Tella/Supporting Files/en.lproj/Localizable.strings
```diff
@@ -377,3 +377,12 @@ This feedback is anonymous, so make sure to include contact information if you w
 "Settings_SettFeedback_ExitFeedback_SheetExpl" = "You can come back anytime to submit your feedback.";
 "Settings_SettFeedback_ExitFeedback_Exit_SheetAction" = "Exit without saving";
 "Settings_SettFeedback_ExitFeedback_Save_SheetAction" = "Save draft";
+
```
dhekra-rouatbi, line 380: I think that it's better to add the english strings in all the languages until we have the translated strings

dhekra-rouatbi (changes requested):

Tella/Scenes/Uwazi/Views/Template/AddTemplatesView.swift
```diff
@@ -30,23 +30,12 @@ struct AddTemplatesView: View {
         .navigationBarTitle("", displayMode: .inline)
         .toolbar {
             LeadingTitleToolbar(title: LocalizableUwazi.uwaziAddTemplateTitle.localized)
-            reloadTemplatesButton()
+            reloadButton(action: { self.uwaziTemplateViewModel.getTemplates() })
```
dhekra-rouatbi, line 33: https://github.com/Horizontal-org/Tella-iOS/pull/119/files#r1505777462

dhekra-rouatbi (changes requested):

Tella/Supporting Files/en.lproj/Localizable.strings
ValbuenaG: Yeah i was going to add it at the end of the code review to avoid merge conflicts if we merge some translations in the meantime.

Tella/Scenes/Resources/ViewModel/ResourcesViewModel.swift
ValbuenaG: I separate the view models for clarity and modularity but makes sense to keep it in just one. Since I change these view models a lot then the others PRs I will add this change to [resources/downloaded](https://github.com/Horizontal-org/Tella-iOS/pull/124). Does that work for you?

dhekra-rouatbi: Yes sure

ValbuenaG: I've been thinking about this and even though I agree that having one viewModel could simplify the data flow, we should also consider the principle of separation of concerns and modularity. Keeping DownloadedResourcesVM and AvailableResourcesVM separate allows us to encapsulate specific functionalities and responsibilities within distinct components, enhancing the readability and reusability of our code. It also makes it easier to manage, debug, and extend each aspect of resource management independently.

What we could do is to find a middle ground between the 2 alternatives. What I have in mind is for the resourceViewModel to act as the parent and orchestrate the interactions between both child view models, this way we could maintain modularity and also avoid passing downloadedResourcesVM to availableResourcesVM.

Let me know if this solution works for you, or if you still prefer the idea of creating just one view model. I'm open to talk about it!

Tella/Scenes/Resources/Views/Common/ResourceCard.swift
```diff
@@ -0,0 +1,33 @@
...
+    var rightButtonAction: () -> Void
+    
+    var body: some View {
+        VStack(spacing: 0) {
+            HStack {
+                Image("resources.pdf")
+                    .padding()
+                ConnectionCardDetail(title: title, subtitle: serverName)
+                Spacer()
+                MoreButtonView(imageName: rightButtonImage, action: rightButtonAction)
```
dhekra-rouatbi, line 24: I don't think that `MoreButtonView` is reusable component because it's a simple button, but if you see that i't reusable button with image you need to update the name for example : `ImageButtonView`

dhekra-rouatbi (changes requested):

Tella/Utils/Utils.swift
```diff
@@ -37,3 +37,12 @@ extension View {
         AnyView(self)
     }
 }
+
+extension View {
+    func customCardStyle() -> some View {
```
dhekra-rouatbi, line 42: I think it's better to create something like 'CardFrameView' and add the padding to the content  or use  'CardFrameView' and adapt it to be used in all the app 
    
    
    
```
 struct CardFrameView<Content:View>: View {
  
        var content : () -> Content
        
        init(@ViewBuilder content : @escaping () -> Content) {
            self.content = content
        }
        
        var body: some View {
            VStack(spacing: 0) {
                self.content()
                    .padding(.all, 16)
            }.background(Color.white.opacity(0.08))
                .cornerRadius(15)
                .padding(.top, 12)
        }
    }
```

dhekra-rouatbi (changes requested):

Tella/Data/Database/Common/DatabaseConstants.swift
```diff
@@ -79,6 +80,12 @@ struct D {
 
 
     static let ctext = "c_text"
+    
+    //resources
+    static let cFilename = "c_filename"
+    static let cExternalId = "c_external_id"
+    static let cSize = "c_size"
+    static let cVaultFileId = "c_vault_file_id"
```
dhekra-rouatbi, line 88: I think `cVaultFileId` is not used

Tella/Data/Database/ResourceDatabase.swift
```diff
@@ -0,0 +1,95 @@
...
+            return resources
+        } catch {
+            return []
+        }
+    }
+    
+    private func getResource(dictionnary: [String: Any]) -> DownloadedResource {
+        let id = dictionnary[D.cId] as? String
+        let externalId = dictionnary[D.cExternalId] as? String
+        let vaultFileId = dictionnary[D.cVaultFileId] as? String
```
dhekra-rouatbi, line 46: I think `vaultFileId ` is not used

Tella/Data/Database/TellaDataBase.swift
```diff
@@ -31,6 +31,8 @@ class TellaDataBase : DataBase {
             case 2:
                 createTemplateTableForUwazi()
                 createUwaziServerTable()
+            case 3:
```
dhekra-rouatbi, line 34: Don't forget to update the `databaseVersion` to `4`

dhekra-rouatbi (changes requested):

Tella/Data/Database/ResourceDatabase.swift
```diff
@@ -0,0 +1,95 @@
...
+    
+    func getDownloadedResources() -> [DownloadedResource] {
+        
+        var resources: [DownloadedResource] = []
+        
+        do {
+            let responseDict = try statementBuilder.selectQuery(tableName: D.tResource)
+            
+            responseDict.forEach { dict in
+                resources.append(getResource(dictionnary: dict))
```
dhekra-rouatbi, line 34: I think it's better to use `decode` function in Array extension 
see `feedbackDict.decode(Feedback.self) `

dhekra-rouatbi (changes requested):

Tella/Data/Database/ResourceDatabase.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+    
+    func deleteDownloadedResource(resourceId: String) -> Result <Bool, Error> {
+        do {
+            let condition = [KeyValue(key: D.cId, value: resourceId)]
+            
+            try statementBuilder.delete(tableName: D.tResource, primarykeyValue: condition)
+            return .success(true)
+        } catch let error {
+            return .failure(error)
```
dhekra-rouatbi, line 92: I think there is a `debugLog(error)` is missing here

Tella/Data/Database/ResourceDatabase.swift
```diff
@@ -0,0 +1,95 @@
...
+
+import Foundation
+
+extension TellaDataBase {
+    func createResourceTable() {
+        let columns = [
+            cddl(D.cId, D.text, primaryKey: true, autoIncrement: false),
+            cddl(D.cExternalId, D.text),
+            cddl(D.cFilename, D.text),
+            cddl(D.cTitle, D.cTitle),
```
dhekra-rouatbi, line 17: I think the type of `cTitle` is `text`

dhekra-rouatbi (changes requested):

Tella/Data/Database/TellaData.swift
```diff
@@ -125,6 +125,23 @@ class TellaData : ObservableObject {
         return database.getReport(reportId: reportId)
     }
     
+    func getResources() -> [DownloadedResource] {
+        return database.getDownloadedResources()
+    }
+    
+    func addResource(resource: Resource, serverId: Int) throws -> String {
```
dhekra-rouatbi, line 132: I think it's better to return the `Result` type : `Result<String, Error>` and then treat the result

Tella/Data/Database/TellaData.swift
```diff
@@ -125,6 +125,23 @@ class TellaData : ObservableObject {
...
+        
+        switch result {
+        case .success(let resourceId):
+            return resourceId
+        case .failure(let error):
+            throw error
+        }
+    }
+    func deleteDownloadedResource(resourceId: String) -> Void {
+        let _ = database.deleteDownloadedResource(resourceId: resourceId)
```
dhekra-rouatbi, line 143: It's better to return the bool result  `Result<Bool,Error>` to treat it in case success or failure

Tella/Data/Networking/APICall/URLRequest.swift
```diff
@@ -12,26 +12,49 @@ typealias APIDataResponse = AnyPublisher<(Data,[AnyHashable:Any]?), APIError>
...
+    func getAPIResponse<Value>(endpoint: any APIRequest) -> APIResponse<Value>
+    where Value: Decodable {
+        fetchData(endpoint: endpoint)
+            .requestJSON()
+            .eraseToAnyPublisher()
+    }
+
+    func getAPIResponseForBinaryData(endpoint: any APIRequest) -> APIResponse<Data> {
+        fetchData(endpoint: endpoint)
+            .tryMap { output in
```
dhekra-rouatbi, line 44: Maybe we can use `.requestData() .eraseToAnyPublisher()` instead of  ` .tryMap ... .mapError ...`

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/Models/Resource.swift
```diff
@@ -0,0 +1,35 @@
...
+    }
+}
+
+struct DownloadedResource: Codable, Identifiable {
+    let id: String
+    let externalId: String
+    let title: String
+    let fileName: String
+    let size: String
+    let serverId: Int?
```
dhekra-rouatbi, line 29: I think it's better to add the `server` instead of  `serverId`
see `func getReports(reportStatus:[ReportStatus]) -> [Report]`  to see the `joinCondition`

dhekra-rouatbi (changes requested):

Tella/Data/Networking/Resources/ResourceService.swift
```diff
@@ -0,0 +1,92 @@
...
+//  Copyright © 2024 HORIZONTAL. All rights reserved.
+//
+
+import Foundation
+import Combine
+
+class ResourceService {
+    private var cancellables: Set<AnyCancellable> = []
+    func getDownloadedResources(from appModel: MainAppModel) -> [DownloadedResourceCardViewModel] {
+        guard let resources = appModel.vaultManager.tellaData?.getResources() else {
```
dhekra-rouatbi, line 15: I suggest this code 
    
```
 func getDownloadedResources(from appModel: MainAppModel) -> [DownloadedResourceCardViewModel] {
        let resources = appModel.vaultManager.tellaData?.getResources()
        let cardsViewModel = resources?.compactMap { resource in
            let selectedServer = appModel.vaultManager.tellaData?.tellaServers.value.first { $0.id == resource.serverId }
            return DownloadedResourceCardViewModel(resource: resource, server: selectedServer)
        }
        return cardsViewModel ?? []
    }
```

and after inscluding the `server` to `DownloadedResource` it will be 

```
 func getDownloadedResources(from appModel: MainAppModel) -> [DownloadedResourceCardViewModel] {
        let resources = appModel.vaultManager.tellaData?.getResources()
        let cardsViewModel = resources?.compactMap { resource in
            return DownloadedResourceCardViewModel(resource: resource)
        }
        return cardsViewModel ?? []
    }

```

dhekra-rouatbi (changes requested):

Tella/Data/Database/ServerDatabase.swift
```diff
@@ -162,6 +162,9 @@ extension TellaDataBase {
             try statementBuilder.delete(tableName: D.tReport,
                                         primarykeyValue: serverCondition)
             
+            try statementBuilder.delete(tableName: D.tResource,
```
ValbuenaG, line 165: @dhekra-rouatbi when a server is disconnected the resources should also be deleted from the device. In this case we should also remove the resource from the file manager, right? Should I create a separate method for this in the serversViewModel?

ValbuenaG: I was thinking of something like this:

```
func deleteServer() {
        guard let serverId = self.currentServer?.id else { return  }

        if currentServer?.serverType == .uwazi {
            mainAppModel.vaultManager.tellaData?.deleteUwaziServer(serverId: serverId)
            
            return
        }
        
        mainAppModel.vaultManager.tellaData?.deleteServer(serverId: serverId)
        let resourcesId = mainAppModel.vaultManager.tellaData?.getResourcesByServerId(serverId: serverId)
        mainAppModel.vaultManager.deleteVaultFile(filesIds: [resourcesId])

    }
    ```
    
    what do you think?

Comment (dhekra-rouatbi) [reply Tella/Data/Database/ServerDatabase.swift:165]: I think it's better to add  `var vaultManager : VaultManagerInterface?`  to `TellaData` and add this code to  `TellaData` (see `class VaultFilesManager : VaultFilesManagerInterface `)

```
    func deleteServer(server:Server) {
        guard let serverId = server.id else { return  }
        
        if server.serverType == .uwazi {
            deleteUwaziServer(serverId: serverId)
        } else {
            deleteServer(serverId: serverId)
        }
        
        let resourcesId =  getResourcesByServerId(serverId: serverId)
        mainAppModel.vaultManager.deleteVaultFile(filesIds: [resourcesId])
        deleteDownloadedResource(resourcesId: resourcesId)
    }


```

File: Tella/Data/Networking/APICall/URLRequest.swift [hunk 20]
```diff
@@ -12,26 +12,38 @@ typealias APIDataResponse = AnyPublisher<(Data,[AnyHashable:Any]?), APIError>
...
-    where Value: Decodable {
+    private func fetchData(endpoint: any APIRequest) -> AnyPublisher<(data: Data, response: URLResponse), Error> {
         do {
-            guard (NetworkMonitor.shared.isConnected) else {
+            guard NetworkMonitor.shared.isConnected else {
                 return Fail(error: APIError.noInternetConnection)
                     .eraseToAnyPublisher()
             }
             let request = try endpoint.urlRequest()
-            request.curlRepresentation()
```
dhekra-rouatbi, line 23: I think `curlRepresentation` is useful to debug the  curl in

dhekra-rouatbi (changes requested):

Tella/Data/Networking/Resources/ResourceService.swift
```diff
@@ -0,0 +1,92 @@
...
+            size: resource.size,
+            createdAt: resource.createdAt
+        )
+    }
+
+    func getAvailableResources(appModel: MainAppModel, servers: [TellaServer]) -> AnyPublisher<
+        [ResourceCardViewModel], APIError
+    > {
+        // Logic to fetch available resources
+        let resourceRepository = ResourceRepository()
```
dhekra-rouatbi, line 46: I suggest to return a list of `Ressource domain model` instead of  `ResourceDTO` from `getResourcesByProject`  API and then map them  to `ResourceCardViewModel`

```
        let publishers = servers.map { server in
            resourceRepository.getResourcesByProject(server: server)
                .map { resource in
                    ResourceCardViewModel(resource: resource, server:server)
                }
        }

```

Tella/Data/Networking/Repositories/ResourceRepository.swift
```diff
@@ -0,0 +1,74 @@
...
+//
+//  Created by gus valbuena on 2/7/24.
+//  Copyright © 2024 HORIZONTAL. All rights reserved.
+//
+
+import Foundation
+import Combine
+
+struct ResourceRepository: WebRepository {
+    func getResourcesByProject(server: TellaServer) -> AnyPublisher<[ResourceDTO], APIError> {
```
dhekra-rouatbi, line 13: It's better to return a list of resource domain model instead of a list of `ResourceDTO`

dhekra-rouatbi (changes requested):

Tella/Data/Networking/Resources/ResourceService.swift
```diff
@@ -0,0 +1,92 @@
...
+            .collect()
+            .map { $0.flatMap { $0 } }
+            .eraseToAnyPublisher()
+    }
+
+    func downloadResource(
+        server: TellaServer, resource: Resource,
+        completion: @escaping (Result<Data, APIError>) -> Void
+    ) {
+        let resourceRepository = ResourceRepository()
```
dhekra-rouatbi, line 75: I think it's better to make this call in the view model

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/ViewModel/AvailableResourcesVM.swift
```diff
@@ -0,0 +1,108 @@
...
+import Foundation
+import Combine
+
+class AvailableResourcesVM: ObservableObject {
+    @Published var appModel: MainAppModel
+    @Published var availableResources: [ResourceCardViewModel] = []
+    @Published var downloadedResourcesVM: DownloadedResourcesVM?
+    @Published var isLoadingList: Bool = false
+    @Published var isLoadingDownload: Bool = false
+    @Published var servers: [TellaServer] = []
```
dhekra-rouatbi, line 19: I think servers  is not used externally and we don't need it to be published 
`private var servers: [TellaServer] = []`

Tella/Scenes/Resources/ViewModel/AvailableResourcesVM.swift
```diff
@@ -0,0 +1,108 @@
...
+//  Created by gus valbuena on 2/5/24.
+//  Copyright © 2024 HORIZONTAL. All rights reserved.
+//
+
+import Combine
+import Foundation
+import Combine
+
+class AvailableResourcesVM: ObservableObject {
+    @Published var appModel: MainAppModel
```
dhekra-rouatbi, line 14: I think we don't need `appModel` to be  `Published`

Tella/Scenes/Resources/ViewModel/AvailableResourcesVM.swift
```diff
@@ -0,0 +1,108 @@
...
+    @Published var availableResources: [ResourceCardViewModel] = []
+    @Published var downloadedResourcesVM: DownloadedResourcesVM?
+    @Published var isLoadingList: Bool = false
+    @Published var isLoadingDownload: Bool = false
```
dhekra-rouatbi, line 18: maybe `isDownloading` is better ?

dhekra-rouatbi (changes requested):

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/ViewModel/ResourcesViewModel.swift
dhekra-rouatbi: I my opinion in the current implentation there is an extra work (pass downloadedResourcesVM in parameter and fetchDownloadedResources) also I see that the two view models are related to each other and I don't see that 'DownloadedResourcesVM' is complex to avoid to include it together with `AvailableResourcesVM` 

After merging we will work on this view model to improve and simplify it .

Tella/Scenes/Resources/ViewModel/DownloadedResourcesVM.swift
```diff
@@ -0,0 +1,38 @@
...
+        self.appModel = mainAppModel
+        fetchDownloadedResources()
+    }
+    
+    func fetchDownloadedResources() {
+        downloadedResources = ResourceService().getDownloadedResources(from: appModel)
+    }
+    
+    func deleteResource(resourceId: String) -> Void {
+        self.appModel.vaultManager.deleteVaultFile(filesIds: [resourceId])
```
dhekra-rouatbi, line 28: I think it's better to add the code to tellaData to delete a downloaded resource

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/Views/AvailableResources.swift
```diff
@@ -0,0 +1,45 @@
...
+            }
+            else if viewModel.availableResources.isEmpty {
+                SectionMessage(text: LocalizableResources.resourcesAvailableEmpty.localized)
+            } else {
+                VStack {
+                    SectionMessage(text: LocalizableResources.resourcesAvailableMsg.localized)
+                }
+                ScrollView {
+                    ForEach(viewModel.availableResources) { resource in
+                        ResourceCard(
```
dhekra-rouatbi, line 27: I think it's better to handle the  `DownloadedResourceCardViewModel` and `ResourceCardViewModel` in the view model
we have an example like this card in `Uwazi` module:  `AddTemplateViewModel` `handleRecieveValue` 

    fileprivate func handleRecieveValue(_ templates: [CollectedTemplate]) {
        self.handleTemplateDownload(templates: templates)
        self.templates = templates
        self.isLoading = false
        
        self.templateItemsViewModel = self.templates.map({ collectedTemplate in
            TemplateItemViewModel(template: collectedTemplate,
                                  downloadTemplate: {self.downloadTemplate(template: collectedTemplate)} ,
                                  deleteTemplate: {self.deleteTemplate(template:collectedTemplate)})
        })
    }

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/Views/DownloadedResources.swift
```diff
@@ -0,0 +1,73 @@
...
+    var body: some View {
+        VStack {
+            SectionTitle(text: LocalizableResources.resourcesDownloadedTitle.localized)
+            if viewModel.downloadedResources.isEmpty {
+                SectionMessage(text: LocalizableResources.resourcesDownloadedEmpty.localized)
+            } else {
+                ScrollView {
+                    LazyVStack {
+                        ForEach(viewModel.downloadedResources) { resource in
+                            ResourceCard(isLoading: false,
```
dhekra-rouatbi, line 23: https://github.com/Horizontal-org/Tella-iOS/pull/119#discussion_r1508842863

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/ViewModel/AvailableResourcesVM.swift
```diff
@@ -0,0 +1,108 @@
...
+                        self.isLoadingList = false
+                        Toast.displayToast(message: LocalizableResources.resourcesAvailableErrorMsg.localized)
+                        print("Error: \(error)")
+                    }
+
+                },
+                receiveValue: { resources in
+                    self.isLoadingList = false
+                    resources.forEach { resource in
+                        let isDownloaded = self.downloadedResourcesVM?.downloadedResources.contains(where: { $0.externalId == resource.id })
```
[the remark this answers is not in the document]
dhekra-rouatbi: yea you're right 
maybe we can try this code  if it works ?
```
self.availableResources = resources.filter { (resource) in
self.downloadedResourcesVM?.downloadedResources.contains{$0.externalId == resource.id} ?? false
}

```

Tella/Data/Database/ServerDatabase.swift
ValbuenaG: great! what do you think of this?

```
func deleteServer(server: Server) {
        guard let serverId = server.id else { return }
        
        if server.serverType == .uwazi {
            deleteUwaziServer(serverId: serverId)
            return
        }
        
        deleteTellaServer(serverId: serverId)
        let resourcesId = getResourceByServerId(serverId: serverId)
        vaultManager?.deleteVaultFile(filesIds: resourcesId)
        
    }

  ```


since the resources will be deleted only if it's a tella server, I'm finishing the execution when is of type uwazi. So it only delete the server and it's done. 

then I'm getting the resourcesID and removing the vaultFiles. The resources in the database are being deleted in the same delete tella server query so we do the same as we do with reports.

File: Tella/Data/Database/TellaData.swift (hunk 14 above)
Comment (ValbuenaG) [reply Tella/Data/Database/TellaData.swift]: Do you still think we should return Result<String, Error> after moving all the logic to tellaData? After moving the vaultManager to tellaData I'm not returning anything because I save the file in the success handler of the operation. Same with the delete one

File: Tella/Scenes/Resources/ViewModel/AvailableResourcesVM.swift (hunk 30 above)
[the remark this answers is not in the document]
Comment (ValbuenaG) [reply Tella/Scenes/Resources/ViewModel/AvailableResourcesVM.swift]: I changed it to this code to avoid the loop on resources:

receiveValue: { resources in
                    self.isLoadingList = false
                    let downloadedIds = Set(self.downloadedResources.map { $0.externalId })
                    let newResources = resources.filter { !downloadedIds.contains($0.id) }
                    self.availableResources.append(contentsOf: newResources)

                }

File: Tella/Data/Networking/Resources/ResourceService.swift (hunk 23 above)
Comment (ValbuenaG) [reply Tella/Data/Networking/Resources/ResourceService.swift:75]: why exactly? Don't you think it's better to abstract the networking logic from the viewModel?

File: Tella/Data/Database/ResourceDatabase.swift [hunk 31]
```diff
@@ -0,0 +1,102 @@
...
+            
+            let decodedResources = try responseDict.decode(DownloadedResource.self)
+            
+            return decodedResources
+        } catch {
+            return []
+        }
+    }
+    
+    func getUnsentFeedbacksw() -> [Feedback] {
```
dhekra-rouatbi, line 39: I think `getUnsentFeedbacksw` is unused

dhekra-rouatbi (changes requested):

Tella/Data/Database/ServerDatabase.swift
dhekra-rouatbi: yeah you're right,  I just suggest doing it with `switch case` because it's more clear and we might have more servers in the future

Tella/Scenes/Resources/Models/Resource.swift
ValbuenaG: so, after adding server to the resource class I will have to modify the Server class to be decodable right? 
or should I decoded the server properties manually on the resource class? bringing this up because I already implement the decode function to parse the array when fetching the resources from the db so I'm curious to know what should be the best approach here

Tella/Data/Database/TellaData.swift
ValbuenaG: bringing this back @dhekra-rouatbi. Since we remove all the logic to tella data do we still need to return a Result type?

Tella/Scenes/Resources/Views/AvailableResources.swift
ValbuenaG: Could we talk this on a call on Monday? I think the implementation differs a little bit with the uwazi one. 

Also if I wanted to set the action for the downloadedResources from the viewModel what would be the best way? Since that action is attempting to show a bottomSheet and the convention we have been following was to do that inside the view. 

I could create a different card for both available and downloaded but we will be missing the point of reusing the component

Tella/Data/Database/TellaData.swift
dhekra-rouatbi: What do you think of this ?

    func addResource(resource: Resource, serverId: Int, data: Data) -> Bool {
        let result = database.addDownloadedResource(resource: resource, serverId: serverId)
        
        switch result {
        case .success(let resourceId):
            guard let isSaved = self.vaultManager?.save(data, vaultFileId: resourceId) else {
                return false
            }
            return true
        case .failure :
            return false
        }
    }

Tella/Scenes/Resources/Models/Resource.swift
dhekra-rouatbi: It's better to use the decode function to parse the ressources, so you can try to modify the server to Decodable
In the `DownloadedResource` class  I think ` init(from decoder: Decoder) throws` is unuseful

Tella/Scenes/Resources/Views/AvailableResources.swift
dhekra-rouatbi: The purpose of this change to simplify the RessourceCard call
for example instead of this 
```
ResourceCard(
            isLoading: resource.isLoading,
            title: resource.title,
            serverName: resource.serverName,
            type: .save,
            action: {viewModel.downloadResource(serverName: resource.serverName, resource: Resource(
                id: resource.id, title: resource.title, fileName: resource.fileName, size: resource.size, createdAt: resource.createdAt
            ))}
        )
```

we can have this

```
        ResourceCard(
            resource: resource,
            type: .save
        )

```

for showing alert we can add a variable to the viewmodel shouldShowAlert , let's try it and then we will find a solution to find the selected ressource after confirming in the  alert

Tella/Scenes/Resources/Models/Resource.swift
ValbuenaG: so I change it but i couldn't fin a way to automatically decode the server after getting the query response. When joining both tables I receive a flat object and the DownloadResource class is expecting a nested server object. So I had to decode manually in the DownlodResource model. 

We could try to parse the server inside the query but I think we will over complex the database function and we will be missing the point of using decode to make the call simpler. 

Anyway let me know what do you think and if you know a better way to decode it!

dhekra-rouatbi: I suggest :

```
    func getDownloadedResources() -> [DownloadedResource] {
        
        do {
            let joinCondition = [JoinCondition(tableName: D.tServer,
                                               firstItem: JoinItem(tableName: D.tResource, columnName: D.cServerId),
                                               secondItem: JoinItem(tableName: D.tServer, columnName: D.cServerId))]
            let responseDict = try statementBuilder.selectQuery(tableName: D.tResource, joinCondition: joinCondition)

            let decodedResources = try responseDict.compactMap({ dict in
                let ressource = try dict.decode(DownloadedResource.self)
                let server = try dict.decode(Server.self)
                ressource.server = server
                return ressource
            })
            
            return decodedResources
        } catch {
            return []
        }
    }


```

and 

```
class DownloadedResource: Codable, Identifiable {
    var id: String
    var externalId: String
    var title: String
    var fileName: String
    var size: String
    var createdAt: String
    var server: Server?
    
    enum CodingKeys: String, CodingKey {
        case id = "c_id"
        case externalId = "c_external_id"
        case title = "c_title"
        case fileName = "c_filename"
        case size = "c_size"
        case createdAt = "c_created_date"
    }
}

```


because the dictionnary from the db is a dictionnary with only level so the server won't be recognized

Tella/Scenes/Resources/Views/ResourcePDFView.swift
```diff
@@ -11,16 +11,18 @@ import SwiftUI
 struct ResourcePDFView: View {
     var file: URL
     var resourceTitle: String
+    @State private var navigationBarHidden = false
     var body: some View {
         ZStack {
-            QuickLookView(file: file)
+            PDFKitView(url: file)
```
ValbuenaG, line 17: @dhekra-rouatbi here I added PDFKit because it provides a more flexible approach for customizing the UI. In this case we wanted to hide the navigation bar when the user scrolls down and show it again when it scrolls up. However I wanted to double check if you think it's fine to use it here. I couldn't find a solution using QuickLook for this kind of customization

Tella/Data/Networking/Resources/ResourceService.swift
dhekra-rouatbi: I think it's better to init the DownloadedResourceCardViewModel with resource (and if there are other attributes) to get a simple  `getDownloadedResources` function
and I suggetc to delete the  non used attributes like `size`, `createdAt` 

I suggest `ResourceCardViewModel` that contains all the attributes and will be used in the downloaded and availble resources:

```
class ResourceCardViewModel: Hashable, Identifiable {
    var id: String
    var externalId: String
    var fileName: String
    var title: String
    var serverName: String
    var type: ResourceCardType
    var action: () -> Void
    
    init(resource: Resource, action:() -> Void) {
        self.id = resource.id
        self.externalId = resource.externalId
        self.fileName = resource.fileName
        ...
    }
}
```

Tella/Scenes/Resources/ViewModel/ResourcesViewModel.swift
```diff
@@ -0,0 +1,218 @@
...
+
+        return Publishers.MergeMany(publishers)
+            .collect()
+            .map { $0.flatMap { $0 } } 
+            .eraseToAnyPublisher()
+    }
+
+    func downloadResource(serverName: String, resource: Resource) {
+        toggleIsLoadingResource(id: resource.id)
+        guard let selectedServer = servers.first(where: { $0.name == serverName }) else {
```
dhekra-rouatbi, line 98: I think it's better to use the sever id instead of `name`

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/ViewModel/ResourceCardViewModel.swift
```diff
@@ -0,0 +1,89 @@
...
+//  ResourceCardViewModel.swift
+//  Tella
+//
+//  Created by gus valbuena on 2/2/24.
+//  Copyright © 2024 HORIZONTAL. All rights reserved.
+//
+
+import Foundation
+
+class ResourceCardViewModel : Identifiable {
```
dhekra-rouatbi, line 11: https://github.com/Horizontal-org/Tella-iOS/pull/119#discussion_r1526502270

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/ViewModel/ServersViewModel.swift
```diff
@@ -19,26 +19,20 @@ class ServersViewModel: ObservableObject {
...
+        guard let server = self.currentServer else { return }
         
-        mainAppModel.vaultManager.tellaData?.deleteServer(serverId: serverId)
+        mainAppModel.tellaData?.deleteServer(server: server)
     }
     
     func deleteAllServersConnection() {
-        mainAppModel.vaultManager.tellaData?.deleteAllServers()
-        mainAppModel.vaultManager.tellaData?.deleteAllServers()
+        mainAppModel.tellaData?.deleteAllServers()
```
dhekra-rouatbi, line 35: duplicated code

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/ViewModel/ResourcesViewModel.swift
```diff
@@ -0,0 +1,218 @@
...
+                    debugLog("Error downloading file: \(error)")
+                    Toast.displayToast(message: error.localizedDescription)
+                }
+                self.toggleIsLoadingResource(id: resource.id)
+            }
+        }
+    }
+    
+
+    func fetchResourceFromServer(
```
dhekra-rouatbi, line 122: In my opinion, I see that it will add more code, if you want to make it more readeable and improve it, you can create function to handle errors and another one to handle sucess result

maybe if we can find a solution to convert the Publisher to async/await -> throws, but we still in this case need to add Task, do catch

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/ViewModel/ResourcesViewModel.swift
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    func getAvailableForDownloadResources() {
+        self.isLoadingList = true
+        self.availableResources = []
+
+        fetchAvailableResources()
+            .receive(on: DispatchQueue.main)
+            .sink(
+                receiveCompletion: { completion in
```
dhekra-rouatbi, line 39: I think it's better to separate the failure and success handling into two functions

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/Views/ResourcePDFView.swift
dhekra-rouatbi: honestly i don't have idea about `PDFKit` but if it  provides more flexible approach for customizing the UI, let's try it

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/Views/AvailableResources.swift
```diff
@@ -0,0 +1,65 @@
...
+//
+
+import SwiftUI
+
+struct AvailableResources: View {
+    @EnvironmentObject var viewModel : ResourcesViewModel
+    var body: some View {
+        VStack {
+            SectionTitle(text: LocalizableResources.resourcesAvailableTitle.localized)
+            VStack {
```
dhekra-rouatbi, line 16: I think it works  without `VStack {` ?

Tella/Scenes/Resources/ResourcesView.swift
```diff
@@ -0,0 +1,95 @@
...
+                    }.padding(.top, 18)
+                    
+                    Spacer()
+                }.padding(.horizontal, 18)
+            }
+        }
+        .navigationBarHidden(true)
+    }
+    
+    var resourceHeaderView: some View {
```
dhekra-rouatbi, line 41: maybe we need to unify the navigation bar because some times we use custom view and sometimes we use native navigation bar

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/ResourcesView.swift
ValbuenaG: Do you mean in the whole app? yeah, it's happening because in the new features we have been adding some constraints to the back button. Since this is not impacting this specific feature I would create a task to refactor the general navigation. What do you think?

dhekra-rouatbi: yeah I meant in the whole app, it's a good idea, 
so we can start creating a centralized nav bar in the next feartures and refactor the old nav bars in the future

Tella/Scenes/Resources/ViewModel/ResourcesViewModel.swift
```diff
@@ -0,0 +1,218 @@
...
+    }
+    
+    private func toggleIsLoadingResource(id: String) {
+        if let index = self.availableResources.firstIndex(where: { $0.id == id }) {
+            self.availableResources[index].isLoading = !self.availableResources[index].isLoading
+            self.availableResources = self.availableResources.map { $0 }
+        }
+    }
+    
+    private func showResourceBottomSheet(resourceId: String, resourceTitle: String) {
```
dhekra-rouatbi, line 215: I suggest to use `resource` instead of  `resourceId` and `resourceTitle` and use  `selectedResource` 

    private func showResourceBottomSheet(resource:Resource) {
        self.selectedResource = resource
        self.onShowResourceBottomSheet?()
    }

dhekra-rouatbi (changes requested):

Tella/Data/Networking/Resources/ResourceService.swift
ValbuenaG: so, in this case we assume that external id is present also in the resource class. But currently is not, because external id exists only in our database. Not in the tella server. 

Should we add the external id also in the resource class and add set it as nil for the available resources? Or put the same value as the id?

Tella/Scenes/Resources/ViewModel/ResourcesViewModel.swift
ValbuenaG: we will only use selectedResource for this function. Why add a new variable instead of passing the specific values that we need for showResourceBottomSheet?

dhekra-rouatbi: The `selectedResource` var  will be used by `showResourceBottomSheet`, `showDeleteResourceConfirmationView`

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/Views/DownloadedResources.swift
```diff
@@ -0,0 +1,79 @@
...
+                                      msgText: LocalizableResources.resourcesDownloadRemoveSheetExpl.localized,
+                                      cancelText: LocalizableResources.resourcesDownloadRemoveCancelSheetAction.localized,
+                                      actionText: LocalizableResources.resourccesDownloadRemoveConfirmSheetAction.localized) {
+                viewModel.deleteResource(resourceId: viewModel.selectedResource?.id ?? "")
+                Toast.displayToast(message: "“\(viewModel.selectedResource?.title ?? "")” \(LocalizableResources.resourcesDownloadRemoveToast.localized)")
+            }
+        }
+    }
+    
+    private func navigateToPDFView(resourceId: String, resourceTitle: String) {
```
dhekra-rouatbi, line 69: maybe we can use the `selectedResource` here ?

dhekra-rouatbi (changes requested):

Tella/Scenes/Resources/Views/DownloadedResources.swift
```diff
@@ -0,0 +1,79 @@
...
+    @EnvironmentObject var viewModel: ResourcesViewModel
+    var body: some View {
+        VStack {
+            SectionTitle(text: LocalizableResources.resourcesDownloadedTitle.localized)
+            if viewModel.downloadedResources.isEmpty && viewModel.availableResources.isEmpty {
+                SectionMessage(text: LocalizableResources.resourcesDownloadedSecondMsg.localized)
+            }
+            else if viewModel.downloadedResources.isEmpty {
+                SectionMessage(text: LocalizableResources.resourcesDownloadedEmpty.localized)
+            } else {
```
dhekra-rouatbi, line 22: maybe it's better to have this view in a separate view

dhekra-rouatbi (changes requested):

PR #126, opened 2024-03-06
handle 401 auth error when a private instance is trying to connect

ValbuenaG:

## Type of change

**Description:**

Fixes #issue-number

## Proposed changes 

*
*
*

### DIFF
--- a/Tella/Scenes/Settings/ViewModel/UwaziServerViewModel.swift
+++ b/Tella/Scenes/Settings/ViewModel/UwaziServerViewModel.swift
@@ -78,11 +78,7 @@ class UwaziServerViewModel: ObservableObject {
         if currentServer != nil {
             updateServer()
         } else {
-            if self.isPublicInstance == true {
-                addServer()
-                return
-            }
-            addPrivateInstanceServer()
+            self.isPublicInstance == true ? addServer() : addPrivateInstanceServer()
         }
     }
 
@@ -343,19 +339,11 @@ class UwaziServerViewModel: ObservableObject {
     
     fileprivate func handleCompletionForSettings(_ completion: Subscribers.Completion<APIError>) {
         switch completion {
-        case .finished:
-            debugLog("Finished")
-            self.isLoading = false
+        case .finished: break
         case .failure(let error):
-            debugLog(error)
-            switch error {
-            case .noInternetConnection:
-                Toast.displayToast(message: error.errorDescription ?? error.localizedDescription)
-            default:
-                break
-            }
-            self.isLoading = false
+            Toast.displayToast(message: error.errorDescription ?? error.localizedDescription)
         }
+        self.isLoading = false
     }
 }
 
--- a/Tella/Scenes/Settings/ViewModel/UwaziServerViewModel.swift
+++ b/Tella/Scenes/Settings/ViewModel/UwaziServerViewModel.swift
@@ -78,12 +78,12 @@ class UwaziServerViewModel: ObservableObject {
         if currentServer != nil {
             updateServer()
         } else {
-            self.isPublicInstance == true ? addServer() : addPrivateInstanceServer()
+            self.isPublicInstance == true ? addServer() : checkURL()
         }
     }
 
     func addServer() {
-        let server = UwaziServer(name: self.name,
+        let server = UwaziServer(name: setting?.siteName,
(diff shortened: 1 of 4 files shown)


Tella/Scenes/Settings/ViewModel/UwaziServerViewModel.swift
```diff
@@ -314,5 +323,39 @@ class UwaziServerViewModel: ObservableObject {
...
+            .receive(on: DispatchQueue.main)
+            .sink(receiveCompletion: { completion in
+                self.handleCompletionForSettings(completion)
+            }, receiveValue: { wrapper in
+                self.name = wrapper.siteName
+                self.addServer()
+            }).store(in: &subscribers)
+    }
+    
+    fileprivate func handleCompletionForSettings(_ completion: Subscribers.Completion<APIError>) {
```
dhekra-rouatbi, line 344: What do you think ?

```
    fileprivate func handleCompletionForSettings(_ completion: Subscribers.Completion<APIError>) {
        switch completion {
        case .finished: break
        case .failure(let error):
            Toast.displayToast(message: error.errorDescription ?? error.localizedDescription)
        }
        self.isLoading = false
    }

```

Tella/Scenes/Settings/ViewModel/UwaziServerViewModel.swift
```diff
@@ -78,12 +78,16 @@ class UwaziServerViewModel: ObservableObject {
         if currentServer != nil {
             updateServer()
         } else {
-            addServer()
+            if self.isPublicInstance == true {
```
dhekra-rouatbi, line 81: self.isPublicInstance == true ? addServer() : addPrivateInstanceServer()

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/ViewModel/UwaziServerViewModel.swift
```diff
@@ -314,5 +323,39 @@ class UwaziServerViewModel: ObservableObject {
         self.username = server.username ?? ""
         self.password = server.password ?? ""
     }
+    
+    
+    fileprivate func addPrivateInstanceServer() {
```
dhekra-rouatbi, line 324: maybe we need a significant function name like  `getServerSettings` or `getServerName`

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/ViewModel/UwaziServerViewModel.swift
```diff
@@ -176,6 +178,8 @@ class UwaziServerViewModel: ObservableObject {
             switch error {
             case .noInternetConnection:
                 Toast.displayToast(message: error.errorDescription ?? error.localizedDescription)
+            case .httpCode(401):
```
dhekra-rouatbi, line 177: Could you use `HTTPErrorCodes`  `unauthorized` instead of `401`

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/ViewModel/UwaziServerViewModel.swift
ValbuenaG: but we are using this function to add the private instance. The thing is that we need to fetch the settings first to get the name of the server.  So naming this getServerSettings would be inaccurate because we are also adding a server.

dhekra-rouatbi, line 324: I think it's better to use one function for both `checkURL()` and  `addPrivateInstanceServer` and if we have token it will be included else it will be nil,  it's a duplicated code

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/ViewModel/UwaziServerViewModel.swift
ValbuenaG: yeah you are right. I removed the addPrivateInstanceServer function and create just one flow using checkURL. the flow will be like this:

- user enter URL
- We check the URL to see if we enable public instance or go straight to login
    - if URL is public: show both options -> user login -> select lang -> create server
    - if URL is private: user login -> select lang -> checkURL -> create server

PR #136, opened 2024-04-22
Bugbash/T-iOS-347

Rimktarii:
Change Localizable.strings files


carohadad:
hi @Rimktarii : you don't need to update the "Rename" text in all the languages, just change it in English. Then I will pull the english file to Lokalise and apply translation memory to get the translation to all the languages and push all those to github in a PR.

The english file is the source of thruth for translations, and all the languages ones are pulled from the platform . I can show the full process on friday if you want!!

Rimktarii:
> hi @Rimktarii : you don't need to update the "Rename" text in all the languages, just change it in English. Then I will pull the english file to Lokalise and apply translation memory to get the translation to all the languages and push all those to github in a PR.
> 
> The english file is the source of thruth for translations, and all the languages ones are pulled from the platform . I can show the full process on friday if you want!!

Only the English version is retained ✅

ValbuenaG (approved):

PR #145, opened 2024-05-09
 GDrive auth flow UI

ValbuenaG:

## Type of change

**Description:**

Fixes #issue-number

## Proposed changes 

*
*
*


Tella/Scenes/Settings/ViewModel/GDriveServerViewModel.swift
```diff
@@ -0,0 +1,44 @@
...
+import Foundation
+import GoogleSignIn
+import GoogleAPIClientForREST
+
+class GDriveServerViewModel: ObservableObject {
+    var googleUser:GIDGoogleUser? = nil
+    @Published var sharedDrives: [GTLRDrive_Drive] = []
+    
+    init() {
+        GIDSignIn.sharedInstance.restorePreviousSignIn { [self] user, error in
```
dhekra-rouatbi, line 18: In my opininion, it's better to centralise the google drive functions in a repository and add a call from the view model to avoid redundancy and minimise the changes in case there is an update in the future. What do you think ?

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/ViewModel/GDriveServerViewModel.swift
ValbuenaG: sure! I was thinking of doing something similar to centralize the g-drive functions. In this case we could create the repository for all the g-drive endpoint calls and unify both view models that I created right? so the GDriveServerVM and GDriveAuthVM will be just one

Tella/Application/TellaApp.swift
```diff
@@ -26,6 +27,10 @@ struct TellaApp: App {
                     if value {
                         self.saveData(lockApptype: .finishBackgroundTasks)
                     }
+                }.onAppear {
+                    gDriveAuthViewModel.restorePreviousSignIn()
```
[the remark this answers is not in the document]
ValbuenaG: I added this on the onAppear method to follow the convention from [google doc](https://developers.google.com/identity/sign-in/ios/sign-in#2_attempt_to_restore_the_users_sign-in_state). The idea is to refresh the token every time the user opens the app. 

We could add a condition to only restorePreviousSignIn if the user has a g-drive server connection so we reduce the amount of calls. What do you think about this?

Tella/Scenes/Settings/ViewModel/GDriveServerViewModel.swift
dhekra-rouatbi: ` In this case we could create the repository for all the g-drive endpoint calls` :  yes 
but we can keep the GDriveServerVM and GDriveAuthVM to make the calls between views and the g-drive repository. What do you think ?

Tella/Application/TellaApp.swift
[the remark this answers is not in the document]
dhekra-rouatbi: `We could add a condition to only restorePreviousSignIn if the user has a g-drive server connection` in this case we need to login ?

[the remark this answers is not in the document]
ValbuenaG: no, but we will need a token for submitting reports to drive. the idea of restorePreviousSignIn is to avoid making the user to re-authenticate and don't let the token expire

Tella/Scenes/Settings/Views/Servers/ServerSelectionView.swift
```diff
@@ -48,6 +50,14 @@ struct ServerSelectionView: View {
                                      isValid: .constant(true), action: {
                 selectedServerType = .uwazi
             }).padding(EdgeInsets(top: 0, leading: 10, bottom: 0, trailing: 10))
+            TellaButtonView<AnyView>(title: "GOOGLE DRIVE",
```
dhekra-rouatbi, line 53: What do you think if we craete a model with the attributes `Title`, ServerConnectionType  and add an array of this model with the different connections

dhekra-rouatbi: and for the action we can create a switch case ?

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/Views/Servers/ServerSelectionView.swift
```diff
@@ -48,6 +50,14 @@ struct ServerSelectionView: View {
...
+                                     nextButtonAction: .action,
+                                     isOverlay: selectedServerType == .gDrive,
+                                     isValid: .constant(true), action: {
+                gDriveVM.handleSignInButton {
```
dhekra-rouatbi, line 57: I think it's better to add it to the next button action ?

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/Views/Servers/GDrive/SelectDriveConnection.swift
```diff
@@ -0,0 +1,88 @@
...
+    case none
+}
+struct SelectDriveConnection: View {
+    @State var selectedDriveConnectionType: DriveConnectionType = .none
+    @ObservedObject var dGriveServerViewModel: GDriveServerViewModel
+    var body: some View {
+        ContainerView {
+            VStack(spacing: 20) {
+                Spacer()
+                GDriveHeaderView(
```
dhekra-rouatbi, line 22: I checked this component and I found that it's use in many views in Settings so can we make the name more general ?

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/ViewModel/GDriveServerViewModel.swift
ValbuenaG: @dhekra-rouatbi I created GDriveRepository for centralizing all the API calls and interactions with GDriveAPI. Since the GDriveAPI works different than the API we interact (tella web and uwazi) I didn't follow the pattern of those repository so I created a singleton instead. Let me know what do you think

dhekra-rouatbi (changes requested):

dhekra-rouatbi:
There is a sign in pop-up that appears and dismiss after opening the app


https://github.com/Horizontal-org/Tella-iOS/assets/91288060/55aab612-96b6-4754-867a-a779cd6e439a

ValbuenaG:
@dhekra-rouatbi hmm weird. It is not happening in mine. I will delete the app and remove the token to check if it's related to that

> There is a sign in pop-up that appears and dismiss after opening the app
> 
>  RPReplay_Final1716994265.mov

Tella/Scenes/Settings/Views/Servers/GDrive/CreateDriveFolder.swift
```diff
@@ -0,0 +1,44 @@
...
+                    .padding(.vertical, 12)
+                Spacer()
+                bottomView
+            }.padding(EdgeInsets(top: 0, leading: 15, bottom: 0, trailing: 15))
+            .navigationBarHidden(true)
+        }
+    }
+    
+    var bottomView: some View {
+        BottomLockView<AnyView>(isValid: .constant(true),
```
dhekra-rouatbi, line 31: maybe we need to add a condition  (`TextfieldView` is valid) to validate the next button ?

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/Views/Servers/GDrive/CreateDriveFolder.swift
```diff
@@ -0,0 +1,44 @@
...
+    @State var fieldContent : String = ""
+    var body: some View {
+        ContainerView {
+            VStack(spacing: 20) {
+                Spacer()
+                GDriveHeaderView(
+                    title: "Create new folder",
+                    subtitle: "Your reports will be uploaded to a new folder on your Google Drive. Choose a name for this folder here."
+                )
+                TextfieldView(fieldContent: $fieldContent, isValid: .constant(true), shouldShowError: .constant(false), fieldType: .text, placeholder: "Folder name")
```
dhekra-rouatbi, line 21: I think it's better to write the arguments on new lines to make them easier to read (Control - M)
  
                TextfieldView(fieldContent: $fieldContent,
                              isValid: .constant(true),
                              shouldShowError: .constant(false),
                              fieldType: .text,
                              placeholder: "Folder name")

dhekra-rouatbi (approved):

Tella/Scenes/Settings/Views/Servers/GDrive/SelectDriveConnection.swift
```diff
@@ -0,0 +1,88 @@
...
+    
+    var connectionsButtons: some View {
+        VStack {
+            TellaButtonView<AnyView>(
+                title: "USE SHARED DRIVE",
+                nextButtonAction: .action,
+                isOverlay: selectedDriveConnectionType == .shared,
+                isValid: .constant(true),
+                action: { selectedDriveConnectionType = .shared }
+            ).padding(.vertical, 5)
```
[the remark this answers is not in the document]
ValbuenaG: but the point of the padding is adding a space between the both buttons. If we apply it only to the VStack we won't have that space

Tella/Scenes/Settings/Views/Servers/GDrive/CreateDriveFolder.swift
```diff
@@ -0,0 +1,52 @@
...
+    var body: some View {
+        ContainerView {
+            VStack(spacing: 20) {
+                Spacer()
+                ServerConnectionHeaderView(
+                    title: "Create new folder",
+                    subtitle: "Your reports will be uploaded to a new folder on your Google Drive. Choose a name for this folder here."
+                )
+                TextfieldView(fieldContent: $fieldContent,
+                              isValid: .constant(true),
```
dhekra-rouatbi, line 23: Could you try `$isValid` instead of adding `.onChange`

                TextfieldView(fieldContent: $fieldContent,
                              isValid: $isValid,
                              shouldShowError: .constant(false),
                              fieldType: .text,
                              placeholder: "Folder name")
                    .padding(.vertical, 12)

dhekra-rouatbi (changes requested):

Tella/Components/ServerConnectionHeaderView.swift
```diff
@@ -0,0 +1,31 @@
...
+//  Copyright © 2024 HORIZONTAL. All rights reserved.
+//
+
+import SwiftUI
+
+struct ServerConnectionHeaderView: View {
+    var title: String
+    var subtitle: String
+    var body: some View {
+        VStack(spacing: 20) {
```
dhekra-rouatbi, line 15: I think the space is not the same between the icon - title (24) and between the title - description (4) ?

dhekra-rouatbi (changes requested):

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/Views/Servers/GDrive/SelectSharedDrive.swift
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+}
+
+struct DriveCardView: View {
+    var sharedDrive: GTLRDrive_Drive
+    @Binding var selectedDrive: String
+    var body: some View {
+        Button(action: {
+            self.selectedDrive = sharedDrive.identifier ?? ""
```
dhekra-rouatbi, line 48: What do you think if we add the selectedDrive action to the view model ?

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/ViewModel/GDriveAuthViewModel.swift
```diff
@@ -0,0 +1,33 @@
...
+import Foundation
+
+class GDriveAuthViewModel: ObservableObject {
+    private let gDriveRepository: GDriveRepositoryProtocol
+
+    init(repository: GDriveRepositoryProtocol = GDriveRepository.shared) {
+        self.gDriveRepository = repository
+    }
+
+    func handleSignInButton(completion: @escaping (Result<Void, Error>) -> Void) {
```
dhekra-rouatbi, line 18: What do you think of `handleSignIn` ?

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/Views/Servers/AddServer/SuccessLoginView.swift
```diff
@@ -13,7 +28,8 @@ struct SuccessLoginView: View {
     
     @EnvironmentObject private var appViewState: AppViewState
     @State var showNextView : Bool = false
-    
+    var navigateToAction: () -> Void
+    var type: SuccessLoginType = .tella
```
dhekra-rouatbi, line 32: What do you think if we use `ServerConnectionType` instead of creating an other attribute (SuccessLoginType) ?

dhekra-rouatbi: and add `buttonContent` to  `ServerConnectionType` extension, maybe also update a significant name instead of `buttonContent`

dhekra-rouatbi (changes requested):

Tella/Data/Networking/Repositories/GDriveRepository.swift
```diff
@@ -0,0 +1,78 @@
...
+            if let error = error {
+                completion(.failure(error))
+                return
+            } else {
+                completion(.success(()))
+            }
+        }
+    }
+    
+    func restorePreviousSignIn(completion: ((_ user: GIDGoogleUser?) -> Void)? = nil) {
```
dhekra-rouatbi, line 42: What do you think if we return a domain model that contains only the parameters that we need instead of `GIDGoogleUser` ?

dhekra-rouatbi (changes requested):

Tella/Scenes/Settings/Views/Servers/GDrive/SelectSharedDrive.swift
```diff
@@ -0,0 +1,68 @@
...
+        }) {
+            HStack {
+                Text(sharedDrive.name ?? "")
+                    .font(.custom(Styles.Fonts.regularFontName, size: 16))
+                    .foregroundColor(.white)
+                Spacer()
+                if selectedDrive == sharedDrive.identifier {
```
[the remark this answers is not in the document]
ValbuenaG: got it! I will add it on the serverConnection PR to avoid conflicts

Tella/Data/Networking/Repositories/GDriveRepository.swift
ValbuenaG: yeah makes sense. I'm wondering if it would be better to not return the GIDGoogleUser and have it only as a state inside the repository. Because we don't need it outside of the repository actually. We use it to pass it as the authorizer for google drive calls. What do you think of this?

For the `GTLRDrive_Drive` I think it makes more sense!

dhekra-rouatbi: yeah you're right, I think you've resolved that by adding the `googleUser` as a private ?

ValbuenaG: yeah! so now we handle the token inside the repository so the rest of the app doesn't have that responsability

Tella/Data/Networking/Repositories/GDriveRepository.swift
```diff
@@ -0,0 +1,86 @@
...
+class GDriveRepository: GDriveRepositoryProtocol  {
+    static let shared = GDriveRepository()
+    
+    private var googleUser: GIDGoogleUser?
+    
+    private var rootViewController: UIViewController? {
+        return UIApplication.shared.windows.first?.rootViewController
+    }
+    
+    func handleSignIn(completion: @escaping (Result<Void, Error>) -> Void) {
```
dhekra-rouatbi, line 29: I suggest to use `async await` instead of `completion`
`async await` simplifies the code and makes it more readable by allowing you to write asynchronous code in a synchronous style

dhekra-rouatbi (changes requested):

Tella/Data/Networking/Repositories/GDriveRepository.swift
```diff
@@ -0,0 +1,105 @@
...
+protocol GDriveRepositoryProtocol {
+    func handleSignIn() async throws -> Void
+    func restorePreviousSignIn() async throws -> Void
+    func handleUrl(url: URL)
+    func getSharedDrives() -> AnyPublisher<[SharedDrive], Error>
+}
+
+class GDriveRepository: GDriveRepositoryProtocol  {
+    static let shared = GDriveRepository()
+
```
Rimktarii, line 23: What do you think about removing the shared and use DI instead?

Tella/Data/Networking/Repositories/GDriveRepository.swift
```diff
@@ -0,0 +1,105 @@
...
+import Foundation
+import GoogleSignIn
+import GoogleAPIClientForREST
+import Combine
+
+protocol GDriveRepositoryProtocol {
+    func handleSignIn() async throws -> Void
+    func restorePreviousSignIn() async throws -> Void
+    func handleUrl(url: URL)
```
Rimktarii, line 17: maybe use directly : 
```
func handleSignIn() async throws
func restorePreviousSignIn() async throws 
```

Tella/Scenes/Settings/ViewModel/GDriveAuthViewModel.swift
```diff
@@ -0,0 +1,38 @@
...
+        Task {
+            do {
+                try await gDriveRepository.handleSignIn()
+                DispatchQueue.main.async {
+                    completion()
+                }
+            } catch let error {
+                DispatchQueue.main.async {
+                    Toast.displayToast(message: error.localizedDescription)
+                }
```
Rimktarii, line 30: Views should not be displayed in the viewModels: 
Maybe you can try this: 
```
    func handleSignIn() async throws {
        try await gDriveRepository.handleSignIn()
    }

```
and handle the rest in `navigateToGDriveFlow` function inside the file `ServerSelectionView` 
Ill write the proposed method in `ServerSelectionView`

Tella/Scenes/Settings/Views/Servers/ServerSelectionView.swift
```diff
@@ -77,6 +80,12 @@ struct ServerSelectionView: View {
             .environmentObject(serverViewModel)
             .environmentObject(serversViewModel))
     }
+    
+    fileprivate func navigateToGDriveFlow() {
+        gDriveVM.handleSignIn { 
+            navigateTo(destination: SelectDriveConnection( dGriveServerViewModel: GDriveServerViewModel()), title: "Select Google drive")
+        }
+    }
```
Rimktarii, line 89: You can replace it by this instead of using completions :
   ```
 fileprivate func navigateToGDriveFlow()   {
        Task {
            do {
                try await gDriveVM.handleSignIn()
                DispatchQueue.main.async {
                    navigateTo(destination: SelectDriveConnection( dGriveServerViewModel: GDriveServerViewModel()), title: "Select Google drive")
                }
            }
            catch{
                Toast.displayToast(message: error.localizedDescription)
            }
        }
    }

```

File: Tella/Scenes/Settings/Views/Servers/GDrive/CreateDriveFolder.swift [hunk 21]
```diff
@@ -0,0 +1,49 @@
...
+                                nextAction: {
+            navigateTo(destination: SuccessLoginView(navigateToAction: {}, type: .gDrive))
+        })
+
+    }
+}
+
+#Preview {
+    CreateDriveFolder()
+}
```
Rimktarii, line 49: Do we really need the previews (also in the other views)!? the canvas is not working in any preview ❌

Tella/Data/Networking/Repositories/GDriveRepository.swift
ValbuenaG: Yeah that's a good idea. My first approach was to use dependency injection but when I tried to implement it before on resources we talked and we decide to not use it for sake of the code consistency. @dhekra-rouatbi what do you think on implement it for this repository?

Tella/Scenes/Settings/Views/Servers/ServerSelectionView.swift
ValbuenaG: if i'm not mistaken the `navigateTo` methodis part of the view layer and manages the UI navigation. I think we are doing this to maintain a separation of responsibilities, so we keep all the navigation logic within the views.

but I get your point! so we could adjust the method to avoid using completion handlers and also handle the toast from the view:
```
fileprivate func navigateToGDriveFlow() {
    Task {
        do {
            try await gDriveVM.handleSignIn()
            DispatchQueue.main.async {
                navigateTo(
                    destination: SelectDriveConnection(gDriveServerViewModel: GDriveServerViewModel(mainAppModel: mainAppModel)),
                    title: LocalizableSettings.settServerGDrive.localized
                )
            }
        } catch {
            DispatchQueue.main.async {
                Toast.displayToast(message: error.localizedDescription)
            }
        }
    }
}
```

ValbuenaG:
@dhekra-rouatbi @Rimktarii What do you think on applying the rest of the changes on the server connection PR? Yesterday I added the translations and finished the stuff for the server connection so we can just include these changes there to avoid merge conflicts.

Tella/Data/Networking/Repositories/GDriveRepository.swift
ValbuenaG: this is done in this [pr](https://github.com/Horizontal-org/Tella-iOS/pull/147)

ValbuenaG:
> @dhekra-rouatbi @Rimktarii What do you think on applying the rest of the changes on the server connection PR? Yesterday I added the translations and finished the stuff for the server connection so we can just include these changes there to avoid merge conflicts.

Also I created the branch `feature/g-drive` to merge there the different branches

Tella/Data/Networking/Repositories/GDriveRepository.swift
Rimktarii: I think maintaining code consistency will lead to a complex codebase if we follow the previous implementation?

Tella/Scenes/Settings/Views/Servers/ServerSelectionView.swift
Rimktarii: Yes, we should have the state and try catch should be sent from the VM and handle the navigations and Toast appearance  in the View

Tella/Scenes/Settings/Views/Servers/GDrive/SelectDriveConnection.swift
```diff
@@ -0,0 +1,88 @@
...
+    @State var selectedDriveConnectionType: DriveConnectionType = .none
+    @ObservedObject var dGriveServerViewModel: GDriveServerViewModel
+    var body: some View {
+        ContainerView {
+            VStack(spacing: 24) {
+                Spacer()
+                ServerConnectionHeaderView(
+                    title: "Select a Drive to connect to",
+                    subtitle: "You can either connect to an organizational Shared Drive or create a new folder in your personal Drive."
+                )
```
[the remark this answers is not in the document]
ValbuenaG: yeah this is done in the server connection PR. Thanks for the reminder

Tella/Data/Networking/Repositories/GDriveRepository.swift
ValbuenaG: I agree on implementing DI for this but let's discuss it on tomorrow's standup so we take a common decision

Tella/Scenes/Settings/Views/Servers/GDrive/SelectSharedDrive.swift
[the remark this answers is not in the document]
ValbuenaG: added to the server connection PR

Tella/Scenes/Settings/Views/Servers/GDrive/SelectSharedDrive.swift
ValbuenaG: solved in this [commit](https://github.com/Horizontal-org/Tella-iOS/pull/147/commits/485fb87adfed525f40f9a290a24d501ed0e19d50) ✅

PR #147, opened 2024-05-27
G-drive connection and submission

ValbuenaG:

## Type of change

**Description:**

Fixes #issue-number

## Proposed changes 

* create gDrive table with the root folder (selected drive or shared drive)
* create a folder in the gDrive of the user
* add unavailable connections: If user has already connected gDrive it would appear as an unavailable connection
* add localizations


Tella/Scenes/Nextcloud/ViewModels/NextcloudReportViewModel.swift
```diff
@@ -0,0 +1,61 @@
...
+        guard let reportId = report.id else { return }
+        //        let resultDeletion = self.tellaData?.deleteGDriveReport(reportId: reportId)
+        var message = ""
+        //        if case .success = resultDeletion {
+        //            message = String.init(format: LocalizableUwazi.uwaziDeletedToast.localized, entity.title ?? "")
+        //        } else {
+        //            message = LocalizableCommon.commonError.localized
+        //        }
+        
+        self.shouldShowToast = true
```
Rimktarii, line 58: do we need this code?

Tella/Scenes/Settings/Models/DeleteServerTexts.swift
```diff
@@ -9,18 +9,58 @@
...
+    var messageText: String {
+        switch self {
+        case .tella:
+            return "if you delete this server, all draft and submitted forms will be deleted from your device."
+        case .uwazi:
+            return "If you delete this server, all draft and submitted entities will be deleted from your device. Delete anyway?"
+        case .gDrive:
+            return "If you delete this server, all draft and submitted reports will be deleted from your device."
+        case .unknown:
+            return ""
```
Rimktarii, line 40: Can we use the same messageText texts look the same for all the servertypes!

Tella/Scenes/Settings/Models/DeleteServerTexts.swift
```diff
@@ -9,18 +9,58 @@
...
+    }
+
+    var cancelText: String {
+        return "CANCEL"
+    }
+    
+    var actionText: String {
+        return "Delete"
+    }
+}
```
Rimktarii, line 51: We can use text from Localizable

Tella/Scenes/Settings/ViewModel/GDriveAuthViewModel.swift
```diff
@@ -12,22 +12,22 @@ import Combine
 class GDriveAuthViewModel: ObservableObject {
     private let gDriveRepository: GDriveRepositoryProtocol
     private var cancellables = Set<AnyCancellable>()
-
-    init(repository: GDriveRepositoryProtocol = GDriveRepository.shared) {
+    @Published var signInState: ViewModelState<String> = .loaded("")
+    init(repository: GDriveRepositoryProtocol) {
```
Rimktarii, line 16: Maybe using Bool instead of String will be better?

Tella/Scenes/Nextcloud/ViewModels/NextcloudReportViewModel.swift
ValbuenaG: hmm we don't need any of the nextcloud for this iteration. This was just for testing purposes of the new report flow. @dhekra-rouatbi are you ok with removing the nexcloud for this? I think you already have a different version in the branch you are working on right? 

We can also leave it as it is because it is going to be changed with the nextcloud release

Tella/Scenes/Settings/Models/DeleteServerTexts.swift
ValbuenaG: I added localization for this but I will leave a different message for Uwazi just for now to follow exactly the figma designs

Tella/Scenes/CommonConnectionReport/ViewModels/ReportMainViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+import Combine
+
+class ReportMainViewModel: ObservableObject {
+    
+    @Published var draftReportsViewModel : [CommonCardViewModel] = []
+    @Published var outboxedReportsViewModel : [CommonCardViewModel] = []
+    @Published var submittedReportsViewModel : [CommonCardViewModel] = []
+    
+    @Published var selectedCell: Pages = .draft
+
```
Rimktarii, line 19: Maybe changig the name to `selectedPage` is more meaningful than `selectedCell` (it looks like UIKit) and changing the type `Pages` to `ConnectionPage` as it is one page and not an array of pages

Tella/Scenes/CommonConnectionReport/Views/Draft/AddFilesToDraftView.swift
```diff
@@ -1,15 +1,18 @@
...
 
 import SwiftUI
 
-struct AddFilesToDraftView: View {
+struct AddFilesToDraftView<T: ServerProtocol>: View {
     
     @EnvironmentObject var appModel: MainAppModel
     @EnvironmentObject var sheetManager: SheetManager
-    @EnvironmentObject var draftReportVM: DraftReportVM
+    @EnvironmentObject var draftReportVM: DraftMainViewModel<T>
```
Rimktarii, line 16: this should be @StateObject and not @EnvironmentObject and then passed in parameter when it is needed

Tella/Scenes/CommonConnectionReport/Views/Draft/DraftView.swift
```diff
@@ -0,0 +1,288 @@
...
+                                             isValid: $viewModel.isValidDescription,
+                                             shouldShowError: $viewModel.shouldShowError,
+                                             shouldShowTitle: true)
+                    
+                    Spacer()
+                        .frame(height: 24)
+                    
+                    AddFilesToDraftView<T>()
+                        .environmentObject(viewModel)
+
```
Rimktarii, line 124: we should remove  `.environmentObject(viewModel) `and pass the `viewmodel` as a parameter

Tella/Scenes/CommonConnectionReport/Views/Draft/DraftView.swift
```diff
@@ -0,0 +1,288 @@
...
+import SwiftUI
+
+struct DraftView<T: ServerProtocol>: View  {
+    @StateObject var viewModel: DraftMainViewModel<T>
+    
+    @State private var menuFrame : CGRect = CGRectZero
+    @State private var shouldShowMenu : Bool = false
+    
+    @EnvironmentObject var mainAppModel: MainAppModel
+    @EnvironmentObject var sheetManager: SheetManager
```
Rimktarii, line 18: it is better to remove and avoid the usage of  `@EnvironmentObject` in our cod. In this situation we can use `viewModel.mainAppModel`

Tella/Scenes/CommonConnectionReport/Views/ReportMainView.swift
```diff
@@ -0,0 +1,236 @@
...
+
+import Foundation
+import SwiftUI
+
+struct ReportMainView: View {
+    
+    @ObservedObject var reportMainViewModel: ReportMainViewModel
+    @EnvironmentObject var sheetManager: SheetManager
+    @EnvironmentObject var mainAppModel: MainAppModel
+
```
Rimktarii, line 17: Here also we have a duplicated mainAppModel so we can remove this` @EnvironmentObject var mainAppModel: MainAppModel`

Tella/Scenes/GDrive/ViewModel/GDriveViewModel.swift
```diff
@@ -0,0 +1,87 @@
...
+    }
+    
+    override func listenToUpdates() {
+        self.mainAppModel.tellaData?.shouldReloadGDriveReports
+            .receive(on: DispatchQueue.main)
+            .sink { result in
+            } receiveValue: { draftReports in
+                self.getReports()
+            }.store(in: &subscribers)
+    }
```
Rimktarii, line 86: Maybe we can try to use one method it in the `ReportMainViewModel` (we should think about it )

Tella/Scenes/CommonConnectionReport/Views/Draft/DraftView.swift
ValbuenaG: got it, I removed it. Just out of curiosity, why is it better to avoid the usage of `@EnvironmentObject`?

Tella/Scenes/GDrive/ViewModel/GDriveViewModel.swift
ValbuenaG: Yeah I agree with this, we should find a way to reuse this method. Let's add it to a next bugbash, thanks for the suggestion rima!

ValbuenaG: https://app.asana.com/0/1199584685142744/1207967764164507 here is the asana task btw, Let's tackle this on the nextcloud release

Tella/Components/Reports/BaseReportsViewModel.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+//  Created by gus valbuena on 6/27/24.
+//  Copyright © 2024 HORIZONTAL. All rights reserved.
+//
+
+import Foundation
+
+protocol ReportsViewModelProtocol: ObservableObject {
+    var selectedCell: Page { get set }
+}
```
Rimktarii, line 13: Same remark here for the `selectedCell` naming

Tella/Data/VaultManager/VaultManager.swift
```diff
@@ -80,7 +80,7 @@ class VaultManager : VaultManagerInterface, ObservableObject{
 
     func loadVaultFileToURL(file vaultFile: VaultFileDB) -> URL? {
 
-        let tmpFileURL = createTempFileURL(pathExtension: vaultFile.fileExtension)
+        let tmpFileURL = createTempFileURL(fileName: vaultFile.name ,pathExtension: vaultFile.fileExtension)
```
Rimktarii, line 84: I think it is more readable to pass directly the `vaultFile`  as a parameter instead of passing multiple params this keeps the method cleaner and easy to understand

Tella/Domain/Entity/CommonReport/BaseReport.swift
```diff
@@ -0,0 +1,102 @@
...
+protocol BaseReportProtocol: Hashable {
+    var id: Int? { get }
+    var title: String? { get }
+    var description: String? { get }
+    var createdDate: Date? { get }
+    var updatedDate: Date? { get }
+    var status: ReportStatus { get }
+    var reportFiles: [ReportFile]? { get }
+    
+    var getReportDate: String { get }
```
Rimktarii, line 20: Unused space

Tella/Scenes/CommonConnectionReport/ViewModels/OutboxMainViewModel.swift
```diff
@@ -0,0 +1,168 @@
...
+//
+
+import Foundation
+import Combine
+
+class OutboxMainViewModel<T: ServerProtocol>: ObservableObject {
+    
+    var mainAppModel : MainAppModel
+    var reportsViewModel : ReportMainViewModel
+
```
Rimktarii, line 16: Is this viewModel of multiple reports or only one report? If it is only one maybe better to rename it to `reportMainViewModel`

Tella/Components/Reports/BaseReportsViewModel.swift
ValbuenaG: we end up not using this protocol so I'll just remove this file

Tella/Scenes/CommonConnectionReport/ViewModels/OutboxMainViewModel.swift
ValbuenaG: it contains all the reports of the connection. It manages the fetching of all the drafts, outbox and submitted reports and the navigation between the different pages

Rimktarii: Then it is better to rename the class to  `ReportsMainViewModel` to indicate that the view model handles all the reports

Tella/Data/VaultManager/VaultManager.swift
ValbuenaG: Yeah, it makes sense! The createTempFileURL method is reusable and is utilized in various parts of the app. In some instances, we only have the fileName or path instead of the VaultFileDB object, which is why we limit the parameters to these two variables. Modifying the method to accept only VaultFileDB would require applying these changes to other methods as well, extending beyond the scope of this PR.

Tella/Scenes/Reports/Outbox/OutboxDetailsView.swift
```diff
@@ -11,6 +11,8 @@ struct OutboxDetailsView<T: ServerProtocol>: View {
     @EnvironmentObject var mainAppModel : MainAppModel
     @Environment(\.presentationMode) var presentationMode: Binding<PresentationMode>
     @EnvironmentObject private var sheetManager: SheetManager
+    @State private var isButtonDisabled = false
+    private let debounceInterval: TimeInterval = 1
```
ValbuenaG, line 15: @Rimktarii @dhekra-rouatbi I added a 1s delay interval to prevent the crashes. I tested it tapping repeatedly and it's working fine. We could use this for now, let me know what do you think!

Tella/Scenes/Settings/ViewModel/GDriveAuthViewModel.swift
```diff
@@ -12,21 +12,23 @@ import Combine
...
+                    self.signInState = .loaded(nil)
                     completion()
                 }
             } catch let error {
                 DispatchQueue.main.async {
-                    Toast.displayToast(message: error.localizedDescription)
+                    self.signInState = .error(error.localizedDescription)
                 }
             }
         }
```
Rimktarii, line 34: You can replace `DispatchQueue.main.async {` with `@MainActor` 
and the method will look like this: 
   ```
 func handleSignIn(completion: @escaping () -> Void) {
        self.signInState = .loading
        Task { @MainActor in
            do {
                try await gDriveRepository.handleSignIn()
                    self.signInState = .loaded(nil)
                    completion()
            } catch let error {
                    self.signInState = .error(error.localizedDescription)
            }
        }
    }

```

File: Tella/Scenes/Settings/ViewModel/GDriveAuthViewModel.swift [hunk 17]
```diff
@@ -12,21 +12,23 @@ import Combine
 class GDriveAuthViewModel: ObservableObject {
     private let gDriveRepository: GDriveRepositoryProtocol
     private var cancellables = Set<AnyCancellable>()
-
-    init(repository: GDriveRepositoryProtocol = GDriveRepository.shared) {
+    @Published var signInState: ViewModelState<String?> = .loaded(nil)
+    init(repository: GDriveRepositoryProtocol) {
```
Rimktarii, line 16: If we are not using the String in `ViewModelState` we can replace it with a boolean to indicate if the user is signed in or not

ValbuenaG: I get your point but we are using the String to display the Toast of the error:

```
 catch let error {
                DispatchQueue.main.async {
                    self.signInState = .error(error.localizedDescription)
                }
            }
  ```
  
  I could create a custom type just to handle the signInState or to handle cases where the loaded state could be nil but I remember some month ago we talked about using just one generic type for managing states

Comment (ValbuenaG) [reply Tella/Scenes/Settings/ViewModel/GDriveAuthViewModel.swift:16]: e.g for this i could create an enum SignInState:

```
enum SignInState: Equatable {
    case init
    case loading
    case loaded
    case failed(String)
}
```

for initialize:

```
@Published var signInState: SignInState = .notStarted
```

I think this could be a clearer approach but the drawback would be creating a new type only for this

File: Tella/Utils/Localizable/LocalizableSettings.swift [hunk 18]
```diff
@@ -110,6 +120,18 @@ enum LocalizableSettings: String, LocalizableDelegate {
...
+    case GDriveSelectTypeDesc = "Setting_Server_GDrive_Select_Type_Desc"
+    case GDriveSelectTypeShared = "Setting_Server_GDrive_Select_Type_Shared"
+    case GDriveSelectTypePersonal = "Setting_Server_GDrive_Select_Type_Personal"
+    case GDriveSelectTypeMoreInfo = "Setting_Server_GDrive_Select_Type_More_Info"
+    case GDriveSelectSharedDriveToolbar = "Setting_Server_GDrive_Select_Shared_Toolbar"
+    case GDriveCreatePersonalFolderTitle = "Setting_Server_GDrive_Create_Personal_Title"
+    case GDriveCreatePersonalFolderDesc = "Setting_Server_GDrive_Create_Personal_Desc"
+    case GDriveCreatePersonalFolderPlaceholder = "Setting_Server_GDrive_Create_Personal_Placeholder"
+    case GDriveSuccessMessage = "Setting_Server_GDrive_Success_Message"
     // About & Help
```
Rimktarii, line 135: Defining the cases should be started by a lowercase letter e.g:
`gDriveSelectTypeToolbar`

PR #161, opened 2024-06-18
Bugbash/T-iOS-428

dhekra-rouatbi:

## Type of change

**Description:**

Implement EXIF metada deletion/preservation

### DIFF
--- a/Tella/Data/VaultManager/VaultFilesManager.swift
+++ b/Tella/Data/VaultManager/VaultFilesManager.swift
@@ -81,7 +81,7 @@ class VaultFilesManager :ObservableObject, VaultFilesManagerInterface {
         return subject.eraseToAnyPublisher()
     }
     
-    func addVaultFile(fileDetail:VaultFileDetails,filePath: URL, parentId: String?) -> AnyPublisher<BackgroundActivityStatus,Never> {
+    func addVaultFile(fileDetail:VaultFileDetails, parentId: String?) -> AnyPublisher<BackgroundActivityStatus,Never> {
         
         let subject = CurrentValueSubject<BackgroundActivityStatus, Never>(.inProgress)
         
--- a/Tella/Data/VaultManager/VaultFilesManager.swift
+++ b/Tella/Data/VaultManager/VaultFilesManager.swift
@@ -102,27 +102,12 @@ class VaultFilesManager :ObservableObject, VaultFilesManagerInterface {
             }
 
             if isSaved {
-                
                 let result = self.vaultDataBase.addVaultFile(file: fileDetail.file, parentId: parentId)
-
-                switch result {
-                case .success:
-
-                    guard let vaultFile = getVaultFile(id: fileDetail.file.id) else {
-                        subject.send(.failed)
-                        return
-                    }
-                    
-                    shouldReloadFiles.send(true)
-                    
-                    // Delete original file
-                    handleDeletionFiles(importedFiles:[fileDetail.importedFile], deleteOriginal: deleteOriginal)
-                    
-                    subject.send(.completed(vaultFile))
-                    
-                default:
-                    subject.send(.failed)
-                }
+                
+                handleResult(result: result,
+                             fileDetails: fileDetail,
+                             deleteOriginal: deleteOriginal,
+                             subject: subject)
             } else {
                 subject.send(BackgroundActivityStatus.failed)
             }
@@ -131,7 +116,31 @@ class VaultFilesManager :ObservableObject, VaultFilesManagerInterface {
         return subject.eraseToAnyPublisher()
     }
     
-    func getFileDetailsStream(_ importedFiles: [ImportedFile]) -> AsyncStream<VaultFileDetails> {
+    private func handleResult(result:Result<Int,Error>, 
+                              fileDetails:VaultFileDetails,
+                              deleteOriginal:Bool,
+                              subject : CurrentValueSubject<BackgroundActivityStatus, Never> ) {
+        switch result {
+        case .success:
+
+            guard let vaultFile = getVaultFile(id: fileDetails.file.id) else {
+                subject.send(.failed)
+                return
+            }
+            
+            shouldReloadFiles.send(true)
+            
+            // Delete original file
+            handleDeletionFiles(importedFiles:[fileDetails.importedFile], deleteOriginal: deleteOriginal)
+            
+            subject.send(.completed(vaultFile))
+            
+        default:
+            subject.send(.failed)
+        }
+    }
+    
+    private func getFileDetailsStream(_ importedFiles: [ImportedFile]) -> AsyncStream<VaultFileDetails> {
         
         // Init AsyncStream with element type = `VaultFileDetails`
         let stream = AsyncStream(VaultFileDetails.self) { continuation in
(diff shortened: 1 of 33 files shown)


Tella/Utils/Extensions/PHAssetExtension.swift
```diff
@@ -0,0 +1,83 @@
...
+    
+    func getDataFromAsset() async throws -> Data? {
+        
+        return try await withCheckedThrowingContinuation { continuation in
+            
+            PHImageManager.default().requestImageDataAndOrientation(for: self, options: nil) { (data, uti, orientation, info) in
+                
+                guard let data else {
+                    guard let error = info?[PHImageErrorKey] as? Error else {
+                        continuation.resume(throwing: RuntimeError("Unknown error occurred"))
```
ValbuenaG, line 31: what do you think of changing the nested guard statement for a `if let`? I think it could be cleaner this way:

```
if let data = data {
    continuation.resume(returning: data)
} else if let error = info?[PHImageErrorKey] as? Error {
    continuation.resume(throwing: error)
} else {
    continuation.resume(throwing: PHAssetError.unknownError)
}
```

ValbuenaG: We could apply the same in the `getAVAssetUrl()` if we decide to go that way

Tella/Utils/Extensions/DataExtension.swift
```diff
@@ -87,3 +87,72 @@ extension Data {
...
+        var gpsDict: [CFString: Any] = [:]
+        
+        let latitude = location.coordinate.latitude
+        let longitude = location.coordinate.longitude
+        
+        gpsDict[kCGImagePropertyGPSLatitude] = abs(latitude)
+        gpsDict[kCGImagePropertyGPSLatitudeRef] = latitude >= 0 ? "N" : "S"
+        gpsDict[kCGImagePropertyGPSLongitude] = abs(longitude)
+        gpsDict[kCGImagePropertyGPSLongitudeRef] = longitude >= 0 ? "E" : "W"
+        gpsDict[kCGImagePropertyGPSTimeStamp] = location.timestamp
```
ValbuenaG, line 150: is it okay if we just assign the timestamp as a string? or do we need to parse it in an specific format?

Tella/Scenes/Camera/ViewModel/CameraViewModel.swift
```diff
@@ -84,14 +85,27 @@ class CameraViewModel: ObservableObject {
...
-        saveFile(urlFile: url)
+        if currentLocation != nil && isPreserveMetadataOn {
+            let url = mainAppModel.vaultManager.createTempFileURL(pathExtension: FileExtension.heic.rawValue)
+            let isSaved = imageData.save(withLocation: currentLocation, fileURL: url)
+            if isSaved {
+                saveFile(urlFile: url)
+            }
+            
+        } else {
+            let url = mainAppModel.vaultManager.saveDataToTempFile(data: imageData,
```
ValbuenaG, line 99: we are declaring multiple times url here. Why don't we just use an if let statement like this:

```
if let url = mainAppModel.vaultManager.saveDataToTempFile(data: imageData, pathExtension: FileExtension.heic.rawValue) {
        saveFile(urlFile: url)
    } else {
        debugLog("Failed to save image without location metadata.")
    }
    
```

Tella/Data/VaultManager/VaultFilesManagerExtension.swift
```diff
@@ -0,0 +1,148 @@
...
+    }
+    
+    func getModifiedUrlFromURL(importedFile:ImportedFile) async -> URL? {
+        
+        guard let urlFile = importedFile.urlFile else { return nil}
+        
+        if importedFile.shouldPreserveMetadata {
+            return urlFile
+        } else {
+            let type = urlFile.fileType
```
ValbuenaG, line 89: can we add this whole block into a new private function?

Tella/Data/VaultManager/VaultFilesManagerExtension.swift
```diff
@@ -0,0 +1,148 @@
...
+    func updateURL(importedFile:inout ImportedFile) async {
+        
+        do {
+            
+            var url : URL?
+            guard let asset = importedFile.asset else {
+                return
+            }
+            
+            let mediaType = asset.mediaType
```
ValbuenaG, line 24: We can remove this and call `asset.mediaType` directly  in the if condition below.

Tella/Data/VaultManager/VaultFilesManagerExtension.swift
```diff
@@ -0,0 +1,148 @@
...
+            
+            if (importedFile.urlFile != nil && mediaType == .image) {
+                return // Image data already saved into the temp fileURL
+            }
+            
+            switch asset.mediaType {
+            case .image:
+                url = try await getImageUrlFromAsset(importedFile:importedFile)
```
ValbuenaG, line 32: Instead of creating a url variable let's just assign it directly to `importedFile.urlFile`. it's less code :)

ValbuenaG (changes requested):
Code looks great overall! Nice work @dhekra-rouatbi.

Rimktarii (approved):

Tella/Utils/Extensions/PHAssetExtension.swift
dhekra-rouatbi: I did that but then I changed with a `guard let` 🤔

Tella/Utils/Extensions/DataExtension.swift
dhekra-rouatbi: yeah you're right  `gpsDict[kCGImagePropertyGPSTimeStamp]` value should have a string value with date formatter `HH:mm:ss.SSSSSS`

Tella/Data/VaultManager/VaultFilesManager.swift
```diff
@@ -77,45 +85,48 @@ class VaultFilesManager :ObservableObject, VaultFilesManagerInterface {
...
         let subject = CurrentValueSubject<BackgroundActivityStatus, Never>(.inProgress)
-
-        if let  isSaved = self.vaultManager?.save(filePath, vaultFileId: fileDetail.file.id) {
-
-            if isSaved {
+        Task {
+            guard let filePath = await getModifiedURL(importedFile: fileDetail.importedFile) else { subject.send(.failed)
+                return}
+            
+            if let  isSaved = self.vaultManager?.save(filePath, vaultFileId: fileDetail.file.id) {
```
Rimktarii, line 97: Maybe using guard let instead if let will make the code more readable ?

Tella/Data/VaultManager/VaultFilesManager.swift
```diff
@@ -77,45 +85,48 @@ class VaultFilesManager :ObservableObject, VaultFilesManagerInterface {
...
+                        
+                        shouldReloadFiles.send(true)
+                        
+                        // Delete original file
+                        handleDeletionFiles(importedFiles:[fileDetail.importedFile], deleteOriginal: deleteOriginal)
+                        
+                        subject.send(.completed(vaultFile))
+                        
+                    default:
                         subject.send(.failed)
```
Rimktarii, line 117: maybe using failure instead defaullt making more sens to send .failed?

Tella/Data/VaultManager/VaultFilesManagerExtension.swift
dhekra-rouatbi: I also added `private`  to the other functions

Tella/Data/VaultManager/VaultFilesManager.swift
dhekra-rouatbi: ` .failed ` is a `BackgroundActivityStatus`, maybe send a `Fail`
How can we modified with failure

Rimktarii: I meant replace 

> `default: ` by `case .failure:`

PR #165, opened 2024-06-28
Fix server subclassing

dhekra-rouatbi:

## Type of change

**Description:**

Fix server subclassing


Tella/Domain/Entity/GDrive/GDriveServer.swift
```diff
@@ -17,31 +17,21 @@ class GDriveServer: Server {
...
+                   serverType: serverType,
+                   allowMultipleConnections: false)
     }
 
     required init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         self.rootFolder = try container.decode(String?.self, forKey: .rootFolder)
         try super.init(from: decoder)
         self.serverType = .gDrive
-        self.allowMultiple = false
```
ValbuenaG, line 44: We need to set allowMultiple as false here, otherwise, it will appear as nil when fetching the servers from the db. this is because the Server class does not handle the decoding of allowMultiple, since is not on the db we are doing this dynamically according to the type of the connection.

ValbuenaG (changes requested):

Tella/Domain/Entity/GDrive/GDriveServer.swift
dhekra-rouatbi: yeah you're right

PR #173, opened 2024-08-23
Nextcloud

Rimktarii:

## Type of change

**Description:**

Fixes #issue-number

## Proposed changes 

*
*
*


Tella/Data/Database/TellaDataBase.swift
```diff
@@ -44,6 +44,9 @@ class TellaDataBase : DataBase {
                 createGDriveServerTable()
                 createGDriveReportTable()
                 createGDriveReportFilesTable()
+                createNextcloudServerTable()
```
ValbuenaG, line 47: let's move nextcloud methods to a new case statement with version 6.

Don't forget to:
[x] add fallthrough to case 5
[x]  update `databaseVersion` to 7 in `DatabaseConstants`

Tella/Data/Networking/Repositories/Nextcloud/NextcloudRepository.swift
```diff
@@ -0,0 +1,410 @@
...
+                }
+            }
+        }
+    }
+    
+    private func fileExists(serverUrlFileName: String) async throws -> Bool?  {
+        
+        let requestBody =
+        """
+        <?xml version=\"1.0\" encoding=\"UTF-8\"?>
```
ValbuenaG, line 324: could we move this to a `Constants` file?

Tella/Data/Networking/Repositories/GDriveRepository.swift
```diff
@@ -416,6 +414,15 @@ class GDriveDIContainer : DIContainer {
     }
 }
 
+class NextcloudDIContainer : DIContainer {
```
ValbuenaG, line 417: why are we adding this here? could we move it to the nextcloud repository?

ValbuenaG:
I found this edge case. I changed my password from nextcloud, then I tried to re authenticate again but I got unauthenticated error even though the username and password is okay. 

then I found that sometimes nextcloud throttled the sign in 30 seconds when there are multiple incorrect attempts. I think we are sending a time out error during those 30 seconds and the user is not able to re authenticate
https://github.com/user-attachments/assets/e0ad9699-7515-4740-add3-4bde6c5d22d0

ValbuenaG (changes requested):

ValbuenaG:
Great work @Rimktarii and @dhekra-rouatbi. It works excellent 🚀  

My only comment regarding the functional test is the behavior with big files. We are submitting the file in chunks of 1mb but the submission of 1mb chunk takes a around ~2s. So for example a file of 900mb will take around 15 to 30 minutes to upload. i know the ideal case for this is doing the submission on background which is not on the scope of this, but could we test how it works if we increase the chunk size for bigger files? let's say 20mb for files bigger than 100mb.

We can also leave this as an improvement for the DRL extension. Just wanted to point this out

Tella/Data/Networking/Repositories/Nextcloud/NextcloudRepository.swift
```diff
@@ -0,0 +1,410 @@
...
+    case progress(progressInfo: NextcloudUploadProgressInfo)
+    case createReport
+    case descriptionSent
+    case nameUpdated(newName:String)
+    case folderRecreated
+}
+
+class NextcloudRepository: NextcloudRepositoryProtocol {
+    
+    private let kRemotePhpFiles = "remote.php/dav/files/"
```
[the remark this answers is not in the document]
Rimktarii: Since there is no official documentation available for the project, we pulled and examined the source code to gather this information

Tella/Data/Database/TellaDataBase.swift
Rimktarii: Changing the databaseVersion to 6 should be done when we provide the production version

Tella/Data/Networking/Repositories/GDriveRepository.swift
Rimktarii: The DIContainer is removed

Tella/Data/Networking/APICall/APIError.swift
```diff
@@ -33,8 +35,27 @@ extension APIError: LocalizedError {
             return LocalizableSettings.settServerNoTokenPresent.localized
         case .driveApiError(let error):
             return customDriveErrorMessage(error: error)
+        case let .nextcloudError(code):
+            return customNcErrorMessage(errorCode: code)
+        case .errorOccured :
+            return LocalizableCommon.commonError.localized
```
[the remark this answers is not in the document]
Rimktarii: Why do you think it is out of the scope of the PR? For my opinion, I think it is not out of the scope as we are displaying this to the user when there is no more significant message to display. 
And for the localizableCommon we can also move it to localizableError

Tella/Data/Networking/Repositories/Nextcloud/NextcloudRepository.swift
```diff
@@ -0,0 +1,410 @@
...
+        
+        if let server {
+            self.setUp(server:server)
+        }
+        
+        guard let rootFolderURL else { throw APIError.errorOccured }
+        
+        let fullURL = rootFolderURL + folderName
+        
+        debugLog(fullURL)
```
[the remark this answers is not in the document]
Rimktarii: yes, we are using it for debugging

Tella/Data/Networking/Repositories/Nextcloud/NextcloudRepository.swift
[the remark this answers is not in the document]
ValbuenaG: sorry i think the reference link is broke. I mean to move this into a constants file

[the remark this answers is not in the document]
Rimktarii: I think it is better to keep it inside the repository as it is a private attribute and only used there, and we don't need it to be outside the Repo

Tella/Scenes/GDrive/Views/GdriveReportMainView.swift
```diff
@@ -0,0 +1,35 @@
...
+//  Copyright © 2024 HORIZONTAL. All rights reserved.
+//
+
+import SwiftUI
+struct GdriveReportMainView: View {
+    
+    @ObservedObject var reportsMainViewModel: GDriveViewModel
+    
+    var body: some View {
+        ReportMainView(reportsMainViewModel: reportsMainViewModel) { id in
```
ValbuenaG, line 15: Don't forget to implement the changes we talked about [here](https://github.com/Horizontal-org/Tella-iOS/pull/183/files/c65f4bef0799d23399e93f69ed5ed4c0131ed412#diff-08c42b92e2cde496f02dfa7c08f7d1477457bc4de31b0c119cbd85b7bfc1ef79). Probably you are working on this but I'll add this comment just to not forget it

ValbuenaG:
Screenshot 2024-09-13 at 3 21 26 PM

I just noticed this UI bug on small-screen devices. We should add an scroll view on `ServerSelectionView`

ValbuenaG:
also the edit button on connections is not working, I'm getting this error:
 
`Publishing changes from within view updates is not allowed, this will cause undefined behavior.`

PR #176, opened 2024-08-28
Select server view : Fix next button action

dhekra-rouatbi:

## Type of change

**Description:**

### DIFF
--- a/Tella/Scenes/Settings/Views/Servers/ServerSelectionView.swift
+++ b/Tella/Scenes/Settings/Views/Servers/ServerSelectionView.swift
@@ -12,7 +12,6 @@ struct ServerSelectionView: View {
     @EnvironmentObject var serversViewModel : ServersViewModel
     @StateObject var serverViewModel : ServerViewModel
     @EnvironmentObject var mainAppModel : MainAppModel
-    @State var selectedServerType: ServerConnectionType? = nil
     @ObservedObject var gDriveVM: GDriveAuthViewModel
     @ObservedObject var gDriveServerVM: GDriveServerViewModel
     @Environment(\.presentationMode) var presentationMode: Binding<PresentationMode>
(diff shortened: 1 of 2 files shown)


Tella/Scenes/Settings/Views/Servers/ServerSelectionView.swift
```diff
@@ -95,14 +94,14 @@ struct ServerSelectionView: View {
     }
     
     fileprivate func navigateToGDriveFlow() {
-        serversViewModel.shouldHideNextButton = true
+        serversViewModel.shouldEnableNextButton = false
```
ValbuenaG, line 97: could we add here:

`serversViewModel.selectedServerType = nil`

If the user cancel the sign in flow the drive connection would be kept as selected and the next button disabled

ValbuenaG: Screenshot 2024-08-28 at 2 24 14 PM

dhekra-rouatbi: yeah you're right, I have the same issue
What do you think if we add  `serversViewModel.shouldEnableNextButton = true` to 
```
.onReceive(gDriveVM.$signInState){ signInState in
            if case .error(let message) = signInState {
                serversViewModel.shouldEnableNextButton = true
                Toast.displayToast(message: message)
            }
        }
        

```

dhekra-rouatbi: I'll try also to add `serversViewModel.selectedServerType = nil`

dhekra-rouatbi: I fixed the issue by adding a `resetView`  function  
```
fileprivate func resetView() {
        serversViewModel.selectedServerType = nil
        serversViewModel.shouldEnableNextButton = false
    }

```

ValbuenaG (approved):

PR #183, opened 2024-09-09
Nextcloud/subclass views

Rimktarii:

## Type of change

**Description:**

Fixes #issue-number

## Proposed changes 

*
*
*

### DIFF
--- a/Tella/Data/Database/TellaDataBase.swift
+++ b/Tella/Data/Database/TellaDataBase.swift
@@ -47,6 +47,7 @@ class TellaDataBase : DataBase {
                 createNextcloudServerTable()
                 createNextcloudReportTable()
                 createNextcloudReportFilesTable()
+                fallthrough
             default :
                 break
             }
--- a/Tella/Scenes/GDrive/Views/GdriveReportMainView.swift
+++ b/Tella/Scenes/GDrive/Views/GdriveReportMainView.swift
@@ -1,0 +1,35 @@
+//
+//  GdriveReportMainView.swift
+//  Tella
+//
+//  Created by RIMA on 6/9/2024.
+//  Copyright © 2024 HORIZONTAL. All rights reserved.
+//
+
+import SwiftUI
+struct GdriveReportMainView: View {
+    
+    @ObservedObject var reportsMainViewModel: GDriveViewModel
+    
+    var body: some View {
+        ReportMainView(reportsMainViewModel: reportsMainViewModel) { id in
+            var destination : any View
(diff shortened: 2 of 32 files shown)


Tella/Scenes/GDrive/Views/GdriveReportMainView.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+
+import SwiftUI
+struct GdriveReportMainView: View {
+    
+    @ObservedObject var reportsMainViewModel: GDriveViewModel
+    
+    var body: some View {
+        ReportMainView(reportsMainViewModel: reportsMainViewModel) { id in
+            var destination : any View
```
ValbuenaG, line 16: could we change this pattern to a method-based navigation with separate functions? I think is more consistent with our codebase and more readable.

it would be something like this:

```
var body: some View {
        ReportMainView(reportsMainViewModel: reportsMainViewModel,
                       showDraftViewAction: navigateToDraftView,
                       showSubmittedViewAction: navigateToSubmittedView,
                       showOutboxViewAction: navigateToOutboxView)
    }
```
and then each method as a separate function:

```
private func navigateToDraftView(_ id: Int?) {
        let draftVM = GDriveDraftViewModel(mainAppModel: reportsMainViewModel.mainAppModel,
                                           repository: viewModel.gDriveRepository,
                                           reportId: id)
        navigateTo(GDriveDraftView(gDriveDraftVM: draftVM, reportsViewModel: viewModel))
    }
```

ValbuenaG: I think is more readable this way but let me know what do you think!

ValbuenaG (commented):
Excellent work Rima! I just tested and it's working great. 

I left a few comments regarding the syntax of the sub views for ReportMainVIew. Let me know what you think!

Tella/Data/Database/TellaDataBase.swift
```diff
@@ -47,6 +47,7 @@ class TellaDataBase : DataBase {
                 createNextcloudServerTable()
                 createNextcloudReportTable()
                 createNextcloudReportFilesTable()
+                fallthrough
```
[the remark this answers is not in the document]
Rimktarii: yes, based on your comment on Nextcloud PR I added it to avoid forgetting it when we move to case 6. So we can keep it for now it is not a problem

Tella/Scenes/GDrive/Views/GdriveReportMainView.swift
Rimktarii: Yes, I am currently working on that refactoring locally. I need to do some other changes before

PR #198, opened 2024-09-25
Description field should not be mandatory in ANY of the server connec…

dhekra-rouatbi:
- [X] Bug fix *(Fixes an issue)*

Fixes #issue-number

T-iOS-500

### DIFF
--- a/Tella/Scenes/CommonConnectionReport/ViewModels/DraftMainViewModel.swift
+++ b/Tella/Scenes/CommonConnectionReport/ViewModels/DraftMainViewModel.swift
@@ -148,4 +148,8 @@ class DraftMainViewModel: ObservableObject {
         self.saveReport()
     }
     
+    func validateTitleAndDescription() {
+        self.isValidTitle =  self.title.textValidator()
+        self.isValidDescription = self.description.textValidator()
+    }
 }
--- a/Tella/Scenes/CommonConnectionReport/ViewModels/DraftMainViewModel.swift
+++ b/Tella/Scenes/CommonConnectionReport/ViewModels/DraftMainViewModel.swift
@@ -93,8 +93,20 @@ class DraftMainViewModel: ObservableObject {
         
         fillReportVM()
     }
-    
-    func validateReport() {}
+
+    func validateReport() {
+        Publishers.CombineLatest4($server,$isValidTitle, $isValidDescription, $files)
+            .map { server, isValidTitle, isValidDescription, files in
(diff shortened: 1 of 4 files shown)


ValbuenaG:
I'm not sure about the task specs. Should it be mandatory for the submission?

Tella/Scenes/Nextcloud/ViewModels/NextcloudDraftViewModel.swift
```diff
@@ -48,6 +36,13 @@ class NextcloudDraftViewModel: DraftMainViewModel {
...
         
         self.objectWillChange.send()
+        
+        DispatchQueue.main.async {
+            self.isValidTitle =  self.title.textValidator()
+            self.isValidDescription = self.description.textValidator()
+            self.objectWillChange.send()
+        }
+
     }
```
Rimktarii, line 46: Maybe we can create a function in `DraftMainViewModel` and then call it in the subclasses : 
Here an example : 
create this method inside `DraftMainViewModel` 
   ```
 func validateTitleAndDescription() {
        self.isValidTitle =  self.title.textValidator()
        self.isValidDescription = self.description.textValidator()
        self.objectWillChange.send()
    }

```
and call it here :


` self.validateTitleAndDescription()
`

The same for the other connections

File: Tella/Scenes/Nextcloud/ViewModels/NextcloudDraftViewModel.swift [hunk 2]
```diff
@@ -48,6 +36,13 @@ class NextcloudDraftViewModel: DraftMainViewModel {
         }
         
         self.objectWillChange.send()
+
```
Rimktarii, line 39: Maybe we can remove this because it is called  twice

Rimktarii (changes requested):

Tella/Scenes/CommonConnectionReport/ViewModels/DraftMainViewModel.swift
```diff
@@ -136,4 +148,8 @@ class DraftMainViewModel: ObservableObject {
         self.saveReport()
     }
     
+    func validateTitleAndDescription() {
+        self.isValidTitle =  self.title.textValidator()
+        self.isValidDescription = self.description.textValidator()
```
Rimktarii, line 153: I think we missed this here :  `self.objectWillChange.send()`

Rimktarii (changes requested):

PR #220, opened 2024-12-23
T-iOS-517 - submit sceen not scrollable (button inaccessible)

dhekra-rouatbi:

## Type of change

- [X] Bug fix *(Fixes an issue)*

### DIFF
--- a/Tella/Scenes/Uwazi/Views/Entity/SummaryEntityView.swift
+++ b/Tella/Scenes/Uwazi/Views/Entity/SummaryEntityView.swift
@@ -23,18 +23,7 @@ struct SummaryEntityView: View {
     var body: some View {
         ContainerView {
             VStack {
-                templateData
-                
-                UwaziDividerWidget()
-                
-                Spacer()
-                    .frame(height: 20)
-                
-                entityTitle
-                
-                entityContent
-                
-                Spacer()
+                entityContentView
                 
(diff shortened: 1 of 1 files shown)


Tella/Scenes/Uwazi/Views/Entity/SummaryEntityView.swift
```diff
@@ -66,6 +54,25 @@ struct SummaryEntityView: View {
...
+                .frame(height: 20)
+            
+            entityTitle
+            
+            entityFilesView
+            
+            Spacer()
+        }
+    }
+
```
Rimktarii, line 75: what do you think of using **scrollOnOverflow()** modifier so the view will not scroll when it is not needed like this: 
      
```

  VStack {

            templateData
            
            UwaziDividerWidget()
            
            Spacer()
                .frame(height: 20)
            
            entityTitle
            
            entityFilesView
            
            Spacer()
        }.scrollOnOverflow()

```

Rimktarii (changes requested):

Tella/Scenes/Uwazi/Views/Entity/SummaryEntityView.swift
dhekra-rouatbi: good idea 👍

PR #246, opened 2025-02-25
Feature/p2p

Rimktarii:

## Type of change

**Description:**

Fixes #issue-number

## Proposed changes 

*
*
*


Tella/Data/Networking/Repositories/PeerToPeer/HTTPParser.swift
```diff
@@ -0,0 +1,100 @@
...
+                instance.currentHeaderField = nil
+            }
+            return 0
+        }
+        
+        settings.on_body = { parser, at, length in
+            guard let at = at else { return 0 }
+            let instance = Unmanaged<HTTPParser>.fromOpaque(parser!.pointee.data).takeUnretainedValue()
+            let value = String(decoding: UnsafeBufferPointer(start: UnsafeRawPointer(at).assumingMemoryBound(to: UInt8.self), count: length), as: UTF8.self)
+            instance.body += value
```
ValbuenaG, line 63: I'm worried about doing this: instance.body += value did you test it with an image, pdf or any binary file? I think this works fine with common string, json or any UTF-8 data but could lead to a crash or data corruption when dealing with binary files or any other data that is not UTF-8 safe.

For binary files we should use a type Data and do so a safe append of the data into the body instead of a string concatenation.

ValbuenaG: If you want let's leave it as it is If it's working, and let's wait for the upload endpoint implementation to check if this needs to be changed.

Tella/Data/Networking/Repositories/PeerToPeer/PeerToPeerServer.swift
```diff
@@ -0,0 +1,336 @@
...
+        }
+    }
+    
+    private func generateRegisterServerResponse() throws -> P2PServerResponse {
+        guard let body = currentHTTPRequest?.body,
+              let registerRequest = body.decodeJSON(RegisterRequest.self) else {
+            throw HTTPStatusCode.badRequest
+        }
+        
+        debugLog("Register request body: \(body)")
```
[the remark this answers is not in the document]
dhekra-rouatbi: `debugLog` only logs messages in debug mode, but I'll remove it

Tella/Data/Networking/Repositories/PeerToPeer/HTTPParser.swift
dhekra-rouatbi: Yes, I agree.  I'll keep it as is for now and will correct it when implementing the file upload endpoint

